PR #8, opened 2015-09-09
Support Ember-CLI-Mocha users.

### LINKED ISSUE #6: QUnit-flavor tests generated for non qunit users
At the very least, if they're not using qunit, we should warn and not generate a test.

```js
import Ember from 'ember';
import { moduleFor, test }
from 'ember-qunit';

moduleFor('validator:better-confirmation', 'Unit | Validator | better-confirmation', {
  needs: ['validator:messages']
});

test('it works', function(assert) {
  var validator = this.subject();
  assert.ok(validator);
});
```

blimmer:
Fixes #6

### DIFF
--- a/blueprints/validator-test/index.js
+++ b/blueprints/validator-test/index.js
@@ -1,0 +1,51 @@
+var path = require('path');
+
+var existsSync = require('exists-sync');
+var walkSync = require('walk-sync');
+
+module.exports = {
+  description: 'Generates a validator unit test',
+
+  init: function() {
+    if (!this.filesPath) {
+      this.files = this._legacyEmberCLIFilesOverride;
+      this.srcPath = this._legacySrcPathOverride;
+    } else {
+      this.filesPath = this._filesPath;
+    }
+  },
+
+  _filesPath: function() {
+    var type;
+    if ('ember-cli-mocha' in this.project.addonPackages) {
+      type = 'mocha';
+    } else if ('ember-cli-qunit' in this.project.addonPackages) {
+      type = 'qunit';
+    } else {
+      this.ui.writeLine('Couldn\'t determine test style - using QUnit');
+      testStyle = 'qunit';
+    }
+    return path.join(this.path, type + '-files');
+  },
+
+  // The files and srcPath method can be removed once
+  // https://github.com/ember-cli/ember-cli/pull/4837 is live.
+  // Also, remove the npm dependencies on walk-sync and exists-sync when fixing.
+  _legacyEmberCLIFilesOverride: function() {
+    if (this._files) { return this._files; }
+
+    var filesPath = this._filesPath();
+
+    if (existsSync(filesPath)) {
+      this._files = walkSync(filesPath);
+    } else {
+      this._files = [];
+    }
+
+    return this._files;
+  },
+
+  _legacySrcPathOverride: function(file) {
+    return path.resolve(this._filesPath(), file);
+  }
+};
--- a/blueprints/validator-test/qunit-files/tests/unit/validators/__name__-test.js
+++ b/blueprints/validator-test/qunit-files/tests/unit/validators/__name__-test.js
@@ -1,4 +1,3 @@
-import Ember from 'ember';
 import { moduleFor, test } from 'ember-qunit';
 
 moduleFor('validator:<%= dasherizedModuleName %>', 'Unit | Validator | <%= dasherizedModuleName %>', {
--- a/blueprints/validator/files/app/validators/__name__.js
+++ b/blueprints/validator/files/app/validators/__name__.js
@@ -1,8 +1,7 @@
-import Ember from 'ember';
 import BaseValidator from 'ember-cp-validations/validators/base';
 
 export default BaseValidator.extend({
-  validate(value, options, model, attribute) {
+  validate(/*value, options, model, attribute*/) {
     return true;
   }
 });
--- a/blueprints/validator-test/mocha-files/tests/unit/validators/__name__-test.js
+++ b/blueprints/validator-test/mocha-files/tests/unit/validators/__name__-test.js
@@ -1,0 +1,20 @@
+/* jshint expr:true */
+import { expect } from 'chai';
+import {
+  describeModule,
+  it
+} from 'ember-mocha';
+
+describeModule(
+  'validator:<%= dasherizedModuleName %>',
+  '<%= classifiedModuleName %>Validator',
+  {
+    needs: ['validator:messages']
+  },
+  function() {
+    it('works', function() {
+      const validator = this.subject();
+      expect(validator).to.be.ok;
+    });
+  }
+);
--- a/package.json
+++ b/package.json
@@ -62,7 +62,8 @@
   "dependencies": {
     "ember-cli-babel": "^5.0.0",
     "ember-cli-version-checker": "^1.1.4",
-    "rsvp": "^3.0.20"
+    "exists-sync": "0.0.3",
+    "walk-sync": "^0.2.0"
   },
   "ember-addon": {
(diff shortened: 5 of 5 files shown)


blimmer:
Lodash blueprinting is ugly - here's what a default test for validator `foo-bar` looks like:

```js
/* jshint expr:true */
import { expect } from 'chai';
import Ember from 'ember';
import {
  describeModule,
  it
} from 'ember-mocha';

describeModule(
  'validator:foo-bar',
  'FooBarValidator',
  {
    // Specify the other units that are required for this test.
    needs: ['validator:messages']
  },
  function() {
    it('works', function() {
      const validator = this.subject();
      expect(validator).to.be.ok;
    });
  }
);

```

stefanpenner:
it should be possible for us to just choose different files based on the current test runner. That may be nicer then creating branching in the same file.

Thoughts?

blimmer:
Would love to see how to do that - do you have an example?

blimmer:
Reminder for me to sync with potentially changing this to an integration test based on [this convo](https://github.com/offirgolan/ember-cp-validations/pull/10#discussion_r39039430) .

blimmer:
@stefanpenner are you thinking something like [this](https://github.com/ember-cli/ember-cli/blob/master/blueprints/resource/index.js)? Doesn't seem obvious how to override install for choosing a different file per test-type.

blimmer:
@rwjblue do you know any secret sauce on how I could have two distinct flavors of a testing file and ember-cli chooses the right one based on which testing framework you use? I've seen [a lot](https://github.com/ember-cli/ember-cli/blob/master/blueprints/component-test/files/tests/__testType__/__path__/__test__.js) of [examples](https://github.com/minutebase/ember-can/blob/master/blueprints/ability-test/index.js#L25) of people doing it this way, but maybe there's some feature I'm missing?

rwjblue:
Key things: 

* Keep the test blueprint separate from the main one. In this case, it means adding a `validator-test` blueprint which is automatically invoked by `ember generate` command for you if it exists ([here](https://github.com/ember-cli/ember-cli/blob/1bdff397b725a7820eaeb09a120b949db0294c56/lib/tasks/generate-from-blueprint.js#L14)). This allows you to avoid duplicating the validator code while having different types of test blueprints.
* Make the `validator-test` blueprint check if `ember-cli-qunit` or `ember-cli-mocha` is installed ([like this](https://github.com/minutebase/ember-can/blob/ef64960b3c925ea66dc9ec1ba5d07325b37afeae/blueprints/ability-test/index.js#L7-L13))
* Override [`files` function](https://github.com/ember-cli/ember-cli/blob/1bdff397b725a7820eaeb09a120b949db0294c56/lib/models/blueprint.js#L291-L302) in `validator-test` blueprint to detect the test type, and use an appropriate files directory.

This might look something like:

```javascript
// blueprints/validator-test/index.js
var path = require('path');
var existsSync = require('exists-sync');
var walkSync = require('walk-sync');

module.exports = {
  description: 'Generates a validator unit test'

  _getFilesPath: function() {
    var type;
    if ('ember-cli-mocha' in this.project.addonPackages) {
      type = 'mocha';
    } else if ('ember-cli-qunit' in this.project.addonPackages) {
      type = 'qunit';
    }
    return path.join(this.path, type + '-files');
  },

  files: function() {
    if (this._files) { return this._files; }

    var filesPath = this._getFilesPath();

    if (existsSync(filesPath)) {
      this._files = walkSync(filesPath);
    } else {
      this._files = [];
    }

    return this._files;
  },

  srcPath: function(file) {
    return path.resolve(this._getFilesPath(), file);
  }
};
```

At this point, you can have a folder structure like:

```
blueprints/validator/index.js
blueprints/validator/files/**/*

blueprints/validator-test/index.js

blueprints/validator-test/qunit-files/**/*

blueprints/validator-test/mocha-files/**/*
```

---

I much prefer this over the being forced to template the files in JS (via strings) as done [here](https://github.com/minutebase/ember-can/blob/ef64960b3c925ea66dc9ec1ba5d07325b37afeae/blueprints/ability-test/index.js#L25-L61).

rwjblue:
As an aside, I believe that the above customizations could either be moved upstream into ember-cli or just made into a small helper/utility library that folks could use.

rwjblue:
I submitted https://github.com/ember-cli/ember-cli/pull/4837 to make this simpler in future ember-cli versions.

blimmer:
Thanks rwjblue!

blimmer:
This is updated and ready for review.

blueprints/validator/files/app/validators/__name__.js
```diff
@@ -1,8 +1,7 @@
-import Ember from 'ember';
 import BaseValidator from 'ember-cp-validations/validators/base';
 
 export default BaseValidator.extend({
-  validate(value, options, model, attribute) {
+  validate(/*value, options, model, attribute*/) {
     return true;
```
blimmer: Default .jshint on ember-cli projects warns on unused attributes.

blueprints/validator-test/qunit-files/tests/unit/validators/__name__-test.js
```diff
@@ -1,4 +1,3 @@
-import Ember from 'ember';
```
blimmer: Default .jshint on ember-cli projects warns on unused attributes.

blueprints/validator-test/index.js
```diff
@@ -0,0 +1,38 @@
...
+    } else if ('ember-cli-qunit' in this.project.addonPackages) {
+      type = 'qunit';
+    }
+    return path.join(this.path, type + '-files');
+  },
+
+  // The files and srcPath method can be removed once
+  // https://github.com/ember-cli/ember-cli/pull/4837 is live.
+  // Also, remove the npm dependencies on walk-sync and exists-sync when fixing.
+  files: function() {
```
rwjblue: Can we future proof this, and only override if on ember-cli before 1.13.9? We are already including ember-cli-version-checker for detecting Ember version and adding ember-get-helper when < Ember 2.0.

blueprints/validator-test/index.js
```diff
@@ -0,0 +1,38 @@
+var path = require('path');
+var existsSync = require('exists-sync');
+var walkSync = require('walk-sync');
+
+module.exports = {
+  description: 'Generates a validator unit test',
+
+  _getFilesPath: function() {
```
rwjblue: Let's name this filesPath to be in line with future versions once that PR lands.

blueprints/validator-test/index.js
rwjblue: In that case, we would detect in init if we need to use overeidden files and srcPath functions, and if we do we would set them in init. Possibly just via something like this:

    this.files = this._legacyEmberCLIFilesOveride;

stefanpenner: should we sniff that the functionality exists? detecting some method?

rwjblue: Sure, we could check in init if this.filesPath exists and if it does, replace it with our custom one above. If it doesn't then we replace this.files and this.srcPath like I mentioned above.

blueprints/validator-test/index.js
```diff
@@ -0,0 +1,38 @@
...
+    var type;
+    if ('ember-cli-mocha' in this.project.addonPackages) {
+      type = 'mocha';
+    } else if ('ember-cli-qunit' in this.project.addonPackages) {
+      type = 'qunit';
+    }
```
offirgolan: If the app isnt using mocha or qunit, we should either default to qunit or exit out with a message.

blimmer:
Sorry for the delay on this - this is fixed up according to comments.

rwjblue:
:+1: - LGTM

Issue #7, opened 2015-09-09
likely need a readme snippet with and example from the HBS side.


stefanpenner:

offirgolan:
Closing due to #10

Issue #5, opened 2015-09-09
Confirmation validation behavior causes strange UX


blimmer:
it's strange to me that there's a validation error message before the user even starts typing the confirmation.

For example, consider a password reset form:
* The user gets here and needs to fill out their password and do a confirmation
* The types their password the first time
* They see a message that the passwords don't match, but they haven't even entered the confirmation yet

For example:

![](http://zippy.gfycat.com/DecimalTenseAnemonecrab.gif)

What do you think about applying the confirmation to the actual `${foo}Confirmation` property itself instead of the `${foo}` property? That way you can associate the error with the confirmation field if it's wrong.

blimmer:
This is a much better UX, IMO:

![](http://zippy.gfycat.com/PopularImmediateCougar.gif)

I wrote a custom validator to do this in my app and am happy to port it over if this is the direction you guys want to go.

stefanpenner:
ya, the idea is to provide helpers (for the UI) to allow the users to opt into various forms.

blimmer:
1.0.1 fixed this.

Issue #2, opened 2015-09-09
isInvalid helper


blimmer:
Many times, a button is disabled until the model is valid. It would be nice to have an `isInvalid: Ember.computed.not('isValid')` on a validated object.

stefanpenner:
seems very reasonable. r? @offirgolan

offirgolan:
I agree. I'll add this tomorrow

blimmer:
available in 1.0.1

PR #10, opened 2015-09-09
Post release bug fixes

offirgolan:
- [x] #2 isInvalid helper
- [x] #5 Confirmation rework. It should now be generated under it's own property. (@stefanpenner let me know what you think about this one. Check out the models/user.js file for an example)
- [x] #7 Added sample hbs in readme
- [x] Fixed validator test blueprint to work with future version of ember


README.md
```diff
@@ -554,3 +563,19 @@ An alias to the first message in the messages collection.
...
 get(user, 'validations.attrs.username.message')
 ```
+
+## Templating Example ##
+Accessing validation information in your templates is really simple. Oh and don't worry, we know the path is pretty long! There are plans to condense that in the roadmap ahead.
+
+```handlebars
+<form>
+  {{input value=model.username placeholder="Username"}}
+  {{#if model.validations.attrs.username.isInvalid}}
```
stefanpenner: a helper for: `(is-valid model 'username')` might be nice

offirgolan: we could also do `(message-for model 'username')`

offirgolan: or even `(v-get model 'username' 'isValid)`

blimmer: I really like the idea of the combined `v-get` helper. the whole `model.validations.attrs.field.prop` is long to type.

blueprints/validator/files/tests/unit/validators/__name__-test.js
```diff
@@ -1,9 +1,8 @@
 import Ember from 'ember';
-import { moduleFor, test }
-from 'ember-qunit';
+import { moduleFor, test } from 'ember-qunit';
 
 moduleFor('validator:<%= dasherizedModuleName %>', 'Unit | Validator | <%= dasherizedModuleName %>', {
-  needs: ['validator:messages']
+  integration: true
```
blimmer: If you're going to make this an integration test, you should move the blueprint to `blueprints/validators/files/tests/integration/validators/__name__-test.js` to match conventions - also should probably change the description above. That said, I don't know if it needs to be an integration test.

tests/dummy/app/models/user.js
```diff
@@ -35,12 +35,13 @@ var Validations = buildValidations({
...
-      message: 'do not match',
-      attributeDescription: 'Email addresses'
-    }),
+    })
   ],
+  emailConfirmation: validator('confirmation', {
+    on: 'email',
+    message: 'do not match',
+    attributeDescription: 'Email addresses'
+  }),
```
blimmer: Nice, I like this much more.

blueprints/validator/files/tests/unit/validators/__name__-test.js
offirgolan: We ran into some registry/container issues with ember-release and above. Converting if to an integration test helped create the container we needed for the test. If you believe this should be a unit test, I can keep as it was. Might just need some more testing though.

blimmer: I definitely don't have a strong opinion on it, was just curious what necessitated the change. Integration tests are very helpful, though slightly slower.

offirgolan: Gotcha yeah I agree. Let me change the folder name to `integration` for now. In the future if want to change it back to unit, then we can make the switch back.

README.md
offirgolan: Moved convo to #12

blueprints/validator/files/tests/unit/validators/__name__-test.js
stefanpenner: > @stefanpenner thoughts on this?

i prefer strict tests, we were slightly rushed yesterday :P

offirgolan: Alright. Let me do some more testing on this.

offirgolan:
Hmmm seems like everything is passing. @stefanpenner the integration test issues could have been something from ember-try? Going to merge this and create a minor release.

PR #11, opened 2015-09-09
Test against ember.js 1.10 - 1.13 in CI

mike-north:
* Delete coverage info from version control, and gitignore
* Remove unnecessary ember-data npm dependency
* Add ember-try scenarios for 1.10, 1.11, 1.12, 1.13 framework versions


offirgolan:
:+1:  Looks good! Thanks @mike-north

package.json
```diff
@@ -41,7 +41,6 @@
     "ember-cli-moment-shim": "0.6.1",
     "ember-cli-qunit": "1.0.1",
     "ember-cli-uglify": "^1.0.1",
-    "ember-data": "~2.0.0",
```
blimmer: Don't we need this for our dummy app tests?

.travis.yml
```diff
@@ -14,7 +14,10 @@ addons:
     repo_token: 32e4df25574ba02aacb75e6d34b9c667044984df279e20a208884d7e9e9d4a8d
 
 env:
-  - EMBER_TRY_SCENARIO=default
+  - EMBER_TRY_SCENARIO=ember-1.10
+  - EMBER_TRY_SCENARIO=ember-1.11
+  - EMBER_TRY_SCENARIO=ember-1.12
```
blimmer: Do we want to consider some of these (1.10) allowed failures?

package.json
offirgolan: Yes we do. Idk how I missed that

.travis.yml
mike-north: why?

package.json
mike-north: Why? There's already a bower reference to ember-data

offirgolan: I think as of right now its needed in both package.json and bower.json. This change broke the demo app so I had to revert it.

rwjblue: We should probably add a handful of acceptance tests to ensure the dummy app still functions...

rwjblue: Yes, ED must be in both package.json and bower.json to be included.

PR #19, opened 2015-09-10
Provide a more permissive email regex.

### LINKED ISSUE #13: Email validation too restrictive
The current email regex rejects gmail style email aliases (eg `ronco+ember@example.org`).

This can definitely be supported using a custom regex, but I wanted to offer up this alias scenario for consideration as I think it's a pretty common use case.

I think you could go much further if interested.  The specific supported syntax for the local-part of email addresses are a bit tough to divine from the various RFCs, but according to [RFC3696](https://tools.ietf.org/html/rfc3696) the following are also valid characters: `! # $ % & ' * + - / = ?  ^ _ ` . { | } ~`.

I'd be happy to provide a PR if there's interest.

blimmer:
Per the HTML5 standards document on validating email addresses on forms http://www.w3.org/TR/html5/forms.html#e-mail-state-(type=email)
Fixes #13.


blimmer:
NVM freaked out momentarily. I don't have access, but redoing the build on travis will show that the tests pass.

offirgolan:
Reran the build just in case :D all passed so gonna merge.

Issue #12, opened 2015-09-09
v-get helper


offirgolan:
Create a general helper to access properties on the validation object.

- `(v-get model 'username' 'isValid')`
- `(v-get model 'isInvalid')`
- `{{v-get model 'username' 'message'}}`
- `{{#each (v-get model 'messages') as |message|}} {{/each}}`

offirgolan:
I think this helper should just be available for Ember >= 2.0. This way it can just be a wrapper around the implementation for the `get` helper introduced in Ember 2.0. 

```javascript
export default Ember.Helper.helper(function(params, hash) {
  // manipulate params 
  // ..
  return Ember.Templates.helpers.get.apply(this, arguments)
});
````

After some playing around, I dont know if this is possible. Before, you can access already defined helpers via `Ember.Handlebars.helpers`. From the docs, it notes that all current helpers are under Ember.Templates but that namespace doesnt exist from within the app. @rwjblue any suggestion here?

**Comment by rwjblue** [comment]
The `Ember.Templates.helpers` namespace is only used to allow us to group documentation, it is not a "real" thing. In general, you cannot invoke helpers as you used to be able to.

---

My suggestion would be add an AST transformation to transform:

```
{{#if (v-get model 'username.isInvalid')}}
{{/if}}
```

Into:

```
{{#if model.validations.attrs.username.isInvalid}}
{{/if}}
```

This allows us to have nice things, but keep the simplicity in the end.

rwjblue:
This intrigued me, so I implemented a first draft of it in JSBin: http://rwjblue.jsbin.com/rujeji/edit?html,js

You can take a peek at ember-gestures to see how to incorporate this into the node-land precompilation for a project:

* https://github.com/runspired/ember-gestures/blob/7986508eb167adc547571e6e6d0aecf791fe9253/index.js#L17-L25
* https://github.com/runspired/ember-gestures/blob/7986508eb167adc547571e6e6d0aecf791fe9253/htmlbars-plugins/touch-action.js

stefanpenner:
> My suggestion would be add an AST transformation to transform:

Ah, this totally slipped my mind!

offirgolan:
This is great! Thanks for the redirection and the first draft. Ill work on this more when I get the chance

blimmer:
This is really cool. Thanks for the tip @rwjblue

offirgolan:
Alright so after some tinkering, I was able to get almost everything we need.

```handlebars
[...]

PR #26, opened 2015-09-11
v-get helper

offirgolan:
- [x] Write `v-get` to `get` AST Transformation
- [x] Get this working with `ember-get-helper` for Ember 1.10 to 1.13.9
- [x] Test cases
- [x] Test out blueprint with version checker
- [x] Update README

Previous discussion can be found at #12 

If you want to play around with the AST Transformation you can just clone this [JSBin](http://jsbin.com/xuyaye/20/edit?html,js,output)


rwjblue:
Testing with the addon for 1.10 - 1.13 support should only require adding it as a devDep. The ember-get-helper addon already [noops itself when running with 2.0.0](https://github.com/jmurphyau/ember-get-helper/blob/master/index.js#L15) (but it does print a warning which will be annoying).

offirgolan:
@rwjblue I'm actually having some issues getting this to work with `ember-get-helper` in Ember 1.13. Seems like it's due to a deprecation to RegisterHelper.

rwjblue:
> As for the inclusion in package.json, I dont think we should include it if it's not necessary.

I am saying that we include it in our `devDependencies` when testing the library itself under 1.x.  Not that we should add it to `dependencies` (which would have all downstream consumers get a warning)

package.json
```diff
@@ -41,10 +41,12 @@
     "ember-cli-moment-shim": "0.6.1",
     "ember-cli-qunit": "1.0.1",
     "ember-cli-uglify": "^1.0.1",
+    "ember-cli-version-checker": "^1.1.4",
```
rwjblue: For this to be usable in the blueprint, it needs to be included in `dependencies` not `devDependencies`.

offirgolan:
>  I am saying that we include it in our devDependencies when testing the library itself under 1.x. Not that we should add it to dependencies (which would have all downstream consumers get a warning)

The previous commit was using Ember 2.0.1, that's why it wasn't in the dependencies. I reverted back to 1.13.8 and included the ember-get-helper in the devDep.

On a side note, would it be possible to test the AST Transformations via integration testing like a component? If not, ill write up some unit tests.

package.json
offirgolan: :+1:

rwjblue:
Yes, you can test any template snippet via `moduleForComponent` (it doesn't need to be related to a component in any way).

Demo: http://jsbin.com/daloriq/edit?js,output

offirgolan:
@rwjblue @stefanpenner @blimmer this is done from my end. Feel free to review when you get some time so we can merge this baby.

tests/integration/helpers/v-get-test.js
```diff
@@ -0,0 +1,88 @@
+import Ember from 'ember';
+import { moduleForComponent, test } from 'ember-qunit';
+import hbs from 'htmlbars-inline-precompile';
+import '../../helpers/ensure-get-registered';
+
+moduleForComponent('helper:v-get', 'Integration | Helper | v-get', {
+  integration: true,
+  beforeEach() {
+    var model = Ember.Object.create({
```
stefanpenner: i suspect this will work without an ember-object.

stefanpenner:
SGTM. would like @rwjblue's +1 on the transform code though.

rwjblue:
Tests seem to cover all of the various cases in the transform and are passing on all supported versions.

LGTM

blimmer:
Sorry to be late to the party on looking at this. Looks great - the AST stuff is cool and makes it way nicer to work with these long strings.

jasonmit:
Will this work at as a runtime plugin as well?  I worked around this in the past:  https://github.com/emberjs/legacy-bind-attr/blob/master/index.js#L12

rwjblue:
@jasonmit - No, it does not work as a runtime plugin as it stands today. We could absolutely do a similar thing to legacy-bind-attr, but I'd prefer for that to be behind a flag if we do (no need to bloat every users `vendor.js` when 95% of them won't use it).  In legacy-bind-attr we needed it to be enabled by default for backwards compat reasons, we don't really have that issue here...

Issue #30, opened 2015-09-15
ember-cp-validations adds about 193 Kb to the vendor.js file size (minimized)


ivanvotti:
The library including its dependencies can be too heavy for projects that don't use `lodash` and `moment.js`

For example, [ember-validations](https://github.com/dockyard/ember-validations) adds just about 16 Kb to the vendor.js file (minimized).

rwjblue:
#31 addresses the lodash part (though I'm not sure how much of the filesize that is)

ivanvotti:
#31 fixes the problem. It's still possible to make the build 42 Kb lighter, if `moment.js` will be optional.

rwjblue:
Ya, we can do that too I think.

offirgolan:
Moment has been removed as a dependency but will be required if you want to use the date validator. This will be available when 2.0 is released.

PR #31, opened 2015-09-15
Remove lodash requirement.

rwjblue:
Remove lodash requirement.

Addresses #30.

### DIFF
--- a/addon/utils/flatten.js
+++ b/addon/utils/flatten.js
@@ -1,0 +1,15 @@
+export default function flatten(array = []) {
+  let result = [];
+
+  for (var i = 0, l = array.length; i < l; i++) {
+    let item = array[i];
+
+    if (Array.isArray(item)) {
+      result = result.concat(flatten(item));
+    } else {
+      result.push(item);
+    }
+  }
+
+  return result;
+}
--- a/addon/utils/validation-result-collection.js
+++ b/addon/utils/validation-result-collection.js
@@ -4,7 +4,7 @@
  */
 
 import Ember from 'ember';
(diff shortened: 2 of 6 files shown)


rwjblue:
Open questions: should we make the steps above a prepublish step and add `lodash` to the `.gitignore`?

I'm generally not a huge fan of inlining someone else's code into my repos, so I kinda think we do want to do this.  It will mean that we cannot point to master via `"ember-cp-validations": "offirgolan/ember-cp-validations"` in `package.json` though...

stefanpenner:
that looks like a crap ton of code to get flatten :P

rwjblue:
@stefanpenner - I'm game to just do a simple internal version of flatten if y'all want. It was fun playing with the custom lodash builds :smile:

offirgolan:
If we are only using flatten, I vote for just creating an internal version. Also, in response to #30, we can opt out of MomentJS and use js Date instead.

rwjblue:
OK, will update to use an internal flatten

rwjblue:
Updated with a super simple flatten implementation. Please review to make sure it meets our requirements (I added a couple of tests with simple scenarios).

ivanvotti:
Thank you, guys. This change makes my vendor.js file 139 Kb lighter.

Here are some numbers I have for minimized vendor.js:
`ember-cp-validations` itself adds just 12 Kb
all its current dependencies (moment.js) add about 42 Kb

`lodash` added about 139 Kb before this PR

stefanpenner:
@ivanvotti we will woke on money next

Issue #27, opened 2015-09-14
Dependent keys for custom validators


mitchlloyd:
Why not let custom validators use fields on the model at their discretion? This would let me support a lot of more "meta" type validations that I have had to deal with.

Looking at https://github.com/offirgolan/ember-cp-validations/blob/v1.1.0/addon/utils/validations-factory.js#L176-L203 I can see that each built-in validator that needs custom dependent keys is hard-coded. Is a general API for this being discussed?

offirgolan:
I believe we have two options here. The first is to expand the `dependent` validator which currently adds a dependent key on `dependent.isTruelyValid` by also adding `dependent` to the list. 

```javascript
var Validations = buildValidations({
   x: [ validator('x-custom'), validator('dependent', { on: ['foo', 'bar.x']}) ]
});
```

The second, is to expose a public api to add custom dependent keys.

```javascript
var Validations = buildValidations({
   x: validator('x-custom', { dependentKeys: ['foo', 'bar.x'] })
});
```

@blimmer  @stefanpenner thoughts?

mitchlloyd:
@offirgolan One thing to keep in mind is that a validator may be dependent on another validator or a property on the model. For instance, the `dependent` validator uses `on` to specify a path to another validator and the `confirmation` validator uses `on` to specify a path to a `_model` attribute.

One use case I'm thinking of is "make sure this model has least one hasMany model with this property". The workaround today could be to create a computed property on the model:

```
hasAtLeastOneFeaturedPost("posts.@each.isFeatured") {
  return _.some(this.get('posts'), (p) => p.get('isFeatured'));
}
```

And then validate that that is true, but it would be nice to be able to express that with just a validator.

```
var Validations = buildValidations({
  featuredPost: validator('has-some', { collection: 'posts', attribute: 'isFeatured' })
});
```

offirgolan:
Im not sure I follow... 

Wouldn't you be able to achieve the same functionality by doing something like this

```js
var Validations = buildValidations({
  featuredPost: validator(function(value, options, model) {
     return _.some(model.get('posts'), (p) => p.get('isFeatured'));
  }, { dependentKeys: ['posts.@each.isFeatured'] })
});
```

PR #41, opened 2015-09-19
dependentKey common validator options

offirgolan:
For #27


offirgolan:
ready for review :smiley_cat:

tests/unit/build-validation-test.js
```diff
@@ -30,28 +30,28 @@
 
  var container, registry;
 
-function register(fullName, factory) {
-  (registry || container).register(fullName, factory);
-}
+ function register(fullName, factory) {
```
rwjblue: Why was leading whitespace added in front of these lines?

offirgolan: woah no idea. Let me fix that

rwjblue:
I left one small comment, but looks good other than that.

In the future, could we keep the auto-generated doc updates in a separate commit than the main code and docs changes?  It makes reading the diff somewhat more difficult (honestly, it would be nice if we could avoid checking in the generated files but that is a whole other issue).

Issue #40, opened 2015-09-19
Needs Documentation


offirgolan:
This is just a comprehensive todo list on what needs to be documented or just needs more extensive documentation.

- [x] `belongsTo` validator
- [x] `hasMany` validator
- [x] `createErrorMessage` in BaseValidator
- [x] `needs` in unit testing
- [x] Better `v-get` helper 
- [x] Ember.Object container dependency

ngouy:
Is there a way to access of an hasMany record not validated ? 
I mean having access to the attributes of the record record_B_1 that invalidate the record_A_1 where
model_A:
```
```
model_A has 'model_B: validator('has-many')' There is no docs about it

ngouy:
Is there a way to access of an hasMany record not validated ? 
I mean having access to the attributes of the record_B that invalidate a record_A where
model_A:
```
const Validations = buildValidations({
  model_B: validator('has-many'),
});

export default Model.extend(Validations, {
  model_Bs: hasMany('model_B')
)};
```

no docs about it.

Issue #9, opened 2015-09-09
i18n message example


jasonmit:
Spoke to @stefanpenner awhile back about this, and sticking to my word to investigate it.

Will poke at it at some point this week to see how we'll be able to leverage the services of ember-intl || ember-i18n to resolve translation keys and handle formatting with params.

offirgolan:
Yeah that was on my todo list as well. I don't think it should be too difficult.... all error messages are stored in `app/validators/messages` so porting that to ember-intl or ember-i18n should be pretty straight forward

stefanpenner:
> Spoke to @stefanpenner awhile back about this, and

:+1:

jasonmit:
@offirgolan I have the sense it will not be that straight forward because both ember intl and i18n require a container lookup and ember intl handles ICU message syntax where your examples use Ember.String.fmt.

offirgolan:
Accessing the messages in the current context is also done via a container lookup. As for the ICU message syntax, that would be something that we would both need to explore.

blimmer:
A few ideas:

* what do we think about implementing this as a validation-messages service? Either way, we'll need to convert the `utils/messages.js` file to be an Ember Object so we can try to inject the i18n/intl service and pull from them if present. this would be nice because the defaults would be cached in the singleton service.

* we'll need to invalidate the messages on locale change

* I'm thinking each flavor of library will need to handle converting to their desired format of string formatting. this library uses String.fmt, i18n uses curlies and intl uses ICU.

What do y'all think?

offirgolan:
> I'm thinking each flavor of library will need to handle converting to their desired format of string formatting. this library uses String.fmt, i18n uses curlies and intl uses ICU.

I think we may need to step away from String.fmt since it seems like it's being deprecated. We can always move over to using curlies or ICU formatted strings and then create two separate files and choose based on which library is being used.

PR #42, opened 2015-09-22
Refactoring and more docs

offirgolan:
* Some much needed refactoring to both the add-on folder and the tests
* Update docs listed in #40


offirgolan:
Ready for review.

docs/docs/validators/common.md
```diff
@@ -61,3 +60,36 @@ The message function is given the following arguments:
...
+
+The method signature is as follows:
+`createErrorMessage(type, options, value, ...formats) {}`
+
+* `type` (**String**): The error message type
+* `options` (**Object**): The validator options that were defined in the model
+* `value`: The current value being evaluated
+* `formats` (**...Array**): String formatters (%@)
+
+If we extended our default messages to include `uniqueUsername: '"%@" already exists'`, we can use this method to generate our error message.
```
stefanpenner: `FMT` is deprecated, in theory we can include a polyfil for it? It may be a good idea to do that, or explore something else.

One idea is to use es2015 template literals +  a function, something like:

```js
formats: [
   (username) => `uniqueUsername: ${username}`,
   (user)     => `uniqueUsername: ${user} `
]
```

offirgolan: That's just currently in place until we include the i18n solution (#9)  which will replace the string.fmt approach we currently have.

stefanpenner: SG

Issue #33, opened 2015-09-17
Ember.String.fmt is deprecated


indirect:
Inside `createErrorMessage`, the [usage of `Ember.String.fmt`](https://github.com/offirgolan/ember-cp-validations/blob/1c3f32a0c6636fd4a4f3387e1264311c1d472d29/addon/validators/base.js#L105) generates a deprecation warning: 

> Ember.String.fmt is deprecated, use ES6 template strings instead. [deprecation id: ember-string-utils.fmt] See https://babeljs.io/docs/learn-es6/#template-strings for more details.

blimmer:
Yep, we're aware of this and this will likely be resolved by [our i18n solution](https://github.com/offirgolan/ember-cp-validations/issues/9#issuecomment-140155506)

PR #43, opened 2015-09-22
support for i18n solutions

### LINKED ISSUE #9: i18n message example

offirgolan:
* Create extendable hook for ember-i18n solutions
* Removed String.fmt dependency
* Better ability to extend message strings and add new ones
* Updated docs and added tests

Should resolve #9, #33

### DIFF
--- a/addon/validators/base.js
+++ b/addon/validators/base.js
@@ -4,12 +4,12 @@
  */
 
 import Ember from 'ember';
+import Messages from './messages';
 
 const {
   get,
   setProperties,
   isNone,
-  typeOf
 } = Ember;
 
 export default Ember.Object.extend({
@@ -35,7 +35,7 @@ export default Ember.Object.extend({
    * Default error messages. Populated by validators/messages
    * @type {Object}
    */
-  defaultMessages: null,
+  errorMessages: null,
 
   // Private
   /**
@@ -48,16 +48,18 @@ export default Ember.Object.extend({
     this._super(...arguments);
     var container = get(this, 'container');
     var options = get(this, 'options');
-    var defaultMessages = {};
+    var errorMessages = Messages;
 
     if(container) {
       // Since default error messages are stored in app/validators/messages, we have to look it up via the container
-      defaultMessages = container.lookupFactory('validator:messages');
+      errorMessages = container.lookupFactory('validator:messages');
     }
+
+    errorMessages = errorMessages.create();
 
     setProperties(this, {
       options: isNone(options) ? {} : options,
-      defaultMessages
+      errorMessages
     });
   },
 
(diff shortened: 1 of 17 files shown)


offirgolan:
Ready for review!

@jasonmit let me know what you think.

jasonmit:
Looks good.  My idea was slightly different in exposing a  virtual method that would allow a consumer, or and addon, to implement.  This way it can scale for anyone needing to add some custom logic between the validation message creation process.

Sorry on phone so can't go too into detail.

blimmer:
I agree that it might be nice to tell the app (in an initializer?) which i18n solution (if any) you're using, and plug in support for that library. All that would do would be to essentially implement the call before this [fallback logic](https://github.com/offirgolan/ember-cp-validations/pull/43/files#diff-ae771607a0869c29daefee660de55711R47). That way you could easily add support for other i18n solutions without needing to pull it into core.

Good work so far, this seems like a step in the right direction, and perhaps the adapter style could be a future improvement.

jasonmit:
I am just opposed to the idea that ember-cp-validations has knowledge of ember-intl, ember-i18n, or any third party packages API.  The reason being, different versions of ember-intl, and I assume ember-i18n, have different method signatures.  When either of us changes the API, this lib will also break and would require a pull request to fix since there is no npm dep on either, rightfully so, enforcing versioning.

I think exposing just a method that can be overridden/implemented is the way to go.  For example, I would ship `ember-intl-cp-validations` which would reopen and replace/implement this method.  This methods sole purpose is to intercept the error message generation/lookup and return an error msg as a string.  This would allow you to continue to use Ember.string.fmt and for any arbitrary third party addon to implement their own custom logic while keeping ember-cp-validations decoupled.

If you do with this approach, any change would require a major version release since it would be immediately API breaking.

offirgolan:
Ah okay I get what you are saying. So for this PR the getMessageFor would be really simple

```js
getMessageFor(type, context = {}) {
    return this.formatMessage(get(this, type), context);
},
```

But the 2 addons you will make will just overwrite this method? Something like

```js
getMessageFor(type, context = {}) {
    if (intl) {
      let m = intl.exists(key);
      if(!isNone(m)) {
        return intl.formatMessage(m, context);
      }
    }
    return this._super(...arguments); // fallback 
},
```

offirgolan:
As for the `String.fmt`, since it is being deprecated, I think we should move away from it rather than creating a polyfill.

jasonmit:
> But the 2 addons you will make will just overwrite this method? Something like

Yessir

offirgolan:
Okay just removed the i18n logic. If I get the :+1: we can go ahead and merge this so we can build the Ember-Intl and Ember-I18n cp-addons. Once those get released, Ill update docs and we can issue a 2.0.0 release.

jasonmit:
You have my thumbs up for what it's worth

blimmer:
I think this solution is going to be much better - nicely decoupled. :+1:

offirgolan:
Alright Im gonna go ahead and merge this. @jasonmit im gonna create an issue to track the status of the two addons. Let me know if you need any help

Issue #44, opened 2015-09-24
Decoupled i18n Addons


offirgolan:
As discussed in #43 and suggested by @jasonmit, we have decided that it would be better to completely decouple the i18n solutions and leave the work to separate addons that will extend the `message` validator with the logic necessary. 

- [x] Support for Ember-Intl
- [x] Support for Ember-I18n
- [x] Support for attributeDescription

Once these two will be supported we can issue a v2.0 release.

jasonmit:
I can complete these both over the weekend :+1:

offirgolan:
Great! Once you're done just post the links on here and I'll update the docs :smile:

jasonmit:
attributeDescription will cause problems, as I'll need a way of overriding that as well when you're generating the message.

jasonmit:
https://github.com/jasonmit/ember-i18n-cp-validations
https://github.com/jasonmit/ember-intl-cp-validations

both published to npm as 0.0.1.  We'll want to fix `attributeDescription` though as part of this issue.

offirgolan:
Oh man I didnt even think about that. What if I create another method in `messages`

```js
getAttributeDescriptionFor(attribute, options) {
  return options.attributeDescription || 'This Field'
}
```

Then in your translation you can maybe have something like the following

```yaml
errors:
  attributeDescriptions:
    username: 'Username'
```

offirgolan:
Also in regards to your log message, just so I dont forget, the link to the message docs will be wrong once the new docs get published. It will be `http://offirgolan.github.io/ember-cp-validations/docs/messages/index.html`.

jasonmit:
> Oh man I didnt even think about that. What if I create another method in messages

Sounds good :+1:

offirgolan:
All necessary changes are now in master so im gonna go ahead and close this issue.

PR #46, opened 2015-09-25
Make attributeDescription be part of message strings for i18n solutions

offirgolan:
In regards to #44

### DIFF
--- a/addon/validators/base.js
+++ b/addon/validators/base.js
@@ -50,7 +50,7 @@ export default Ember.Object.extend({
     var options = get(this, 'options');
     var errorMessages = Messages;
 
-    if(container) {
+    if (container) {
       // Since default error messages are stored in app/validators/messages, we have to look it up via the container
       errorMessages = container.lookupFactory('validator:messages');
     }
@@ -85,25 +85,24 @@ export default Ember.Object.extend({
    * @param  {Object} context     : Context for string replacement
    * @return {String}             : The generated message
    */
-  createErrorMessage(type, options, value, context) {
+  createErrorMessage(type, options, value, context = {}) {
     options = options || {};
-    var attribute = options.attributeDescription || 'This field';
+    context.attributeDescription = options.attributeDescription;
+
     var messages = this.get('errorMessages');
     var message;
-
-    attribute = `${attribute.trim()} `;
 
     if (options.message) {
       if (typeof options.message === 'string') {
         message = messages.formatMessage(options.message, context);
       } else if (typeof options.message === 'function') {
-        message = options.message.call(this, ...arguments);
+        message = options.message.apply(this, arguments);
         message = isNone(message) ? messages.getMessageFor(type, context) : messages.formatMessage(message, context); // fail-safe to default message of type
       }
     } else {
       message = messages.getMessageFor(type, context);
     }
 
-    return (attribute + message).trim();
+    return message.trim();
   }
 });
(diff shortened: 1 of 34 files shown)


offirgolan:
@jasonmit let me know if this works for you and I'll merge it

addon/validators/base.js
```diff
@@ -87,23 +87,21 @@ export default Ember.Object.extend({
    */
   createErrorMessage(type, options, value, context) {
     options = options || {};
-    var attribute = options.attributeDescription || 'This field';
```
jasonmit: Since you're removing it, you'll need to add the default in your own message hash

jasonmit:
:+1: although I am unsure how well prefixing works across all locales in terms of how many languages construct sentences.

offirgolan:
hmmmm yeah you're totally right. What if get rid of this and instead make attributeDescription part of the actual message context?

So instead of  `invalid: 'is invalid'` it would be `invalid: '{attributeDescription} is invalid'`

That way in the different local `{attributeDescription}` can be places anywhere in the message.

jasonmit:
:+1:

offirgolan:
Sounds good. Im going to be out of town this weekend so I wont get to it until Monday.

offirgolan:
@jasonmit this should do the trick. I also made a property in there `defaultAttributeDescription` which you can overwrite with a CP that grabs `errors.attributeDescriptions.default` from a translation. Up to you :)

addon/validators/base.js
```diff
@@ -85,25 +85,24 @@ export default Ember.Object.extend({
    * @param  {Object} context     : Context for string replacement
    * @return {String}             : The generated message
    */
-  createErrorMessage(type, options, value, context) {
+  createErrorMessage(type, options, value, context = {}) {
     options = options || {};
```
stefanpenner: should we also default args for options?

offirgolan: Yeah I can do that as well.

offirgolan: Actually, doing that would change the method signature which would have to be propagated through each validator and it would also cause a change in the message function feature.

stefanpenner: ah ok

offirgolan: Since we are going to release 2.0, we can justify this change. Up to you

stefanpenner: its such a small thing, I am indifferent. I just wanted to point out the difference.

You should choose based on impact to users in the wild. If it wont cause much (or any grief) go for it, if it causes some. It may not be worth it.

offirgolan: Sounds good. Lets just leave it for now. I don't think we should change two public API's for a default arg.

stefanpenner: :+1:

jasonmit:
Looks good, ping me once versioned and I'll update the micro libs and their examples.

jasonmit:
> I also made a property in there defaultAttributeDescription which you can overwrite with a CP that grabs errors.attributeDescriptions.default from a translation. Up to you :)

I'm not sure what you mean here

jasonmit:
And one other thing, ember-i18n uses double curly to mark dynamic segments of a translation.   May need to adjust your regex to account for that.

offirgolan:
@stefanpenner :snail: ready for review

blimmer:
LGTM! :dancer:

stefanpenner:
:+1:

Issue #38, opened 2015-09-19
attributeDescription should be per-field, not per-validator


lolmaus:
It is burdensome to pass `attributeDescription` to every validator of a field:

```js
const Validations = buildValidations({
  name: [
    validator('presence', {
      presence: true,
      attributeDescription: 'Course name (language)'
    }),
    validator('length', {
      min: 1,
      attributeDescription: 'Course name (language)'
    }),
    validator('no-whitespace-around', {
      attributeDescription: 'Course name (language)'
    })
  ]
});
``` 

It is obvious that conceptually `attributeDescription` is an entity that belongs to the whole field and should be shared among all validators.

I suggest the following API:

```js
/**
 * Top level method that will ultimately return a mixin with all CP validations
 * @param  {Object} validations  Validation rules
 * @param  {Object} commonOptions  Options to be includd into every validator unless defined explicitly
 * @return {Ember.Mixin}
 */
export default function buildValidations(validations) {
```

which would drastically simplify the above example:

```js
const Validations = buildValidations({
  name: [
    validator('presence', true),
    validator('length', { min: 1 }),
    validator('no-whitespace-around')
  ]
},
{
  attributeDescription: 'Course name (language)'
});
```

offirgolan:
Thanks for the suggestion. I definitely had something like this in my head for future implementation. Ill take a look into this more when I have some time.

offirgolan:
This will be available once we release 2.0! Thanks again for the suggestion :smile:

PR #49, opened 2015-09-30
Add hooks in base validator, remove moment dependency, default options

offirgolan:
* Add buildOptions and processOptions hooks
* Default options for validators (#38)
* All options can be functions that are called right before validate
via processOptions
* Removed moment dependency (#30)

### DIFF
--- a/addon/validators/base.js
+++ b/addon/validators/base.js
@@ -8,7 +8,8 @@ import Messages from './messages';
 
 const {
   get,
-  setProperties,
+  set,
+  merge,
   isNone,
 } = Ember;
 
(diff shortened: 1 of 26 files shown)


offirgolan:
@stefanpenner ready for review :rocket:

addon/validators/base.js
```diff
@@ -48,26 +55,57 @@ export default Ember.Object.extend({
     this._super(...arguments);
     var container = get(this, 'container');
     var options = get(this, 'options');
+    var defaultOptions = get(this, 'defaultOptions');
     var errorMessages = Messages;
 
-    if(container) {
+    if (container) {
```
stefanpenner: i prefer this, but other parts of the code don't leave space here. A future PR should apply a uniform style

stefanpenner:
SGTM

PR #55, opened 2015-10-16
Add presence-or validator

elwayman02:
This validator checks for the presence of one or more values based on the attribute tied to the validator and the `dependentKeys` array passed into `options`.

Users can also pass an `exclusive: true` option that makes this an exclusive-or check. That is, it expects the presence of one and only one of the values being watched.

### DIFF
--- a/app/validators/presence-or.js
+++ b/app/validators/presence-or.js
@@ -1,24 +1,40 @@
 import Ember from 'ember';
 import BaseValidator from 'ember-cp-validations/validators/base';
 
-const { isPresent } = Ember;
+const { isBlank, isEqual, isPresent, typeOf } = Ember;
 
 export default BaseValidator.extend({
-  keysPresent(value, keys, model) {
+  keyPresent(model, key) {
+    const value = key ? model.get(key) : model;
+    if (isBlank(value)) {
+      return false;
+    }
+    // If an object is present, it may be an empty `PromiseObject` or `PromiseArray` from an Ember-Data relationship
+    const constructor = value.constructor;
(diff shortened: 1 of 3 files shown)


elwayman02:
@offirgolan thoughts?

offirgolan:
This looks great but I dont know if it would be the right fit to be part of the pre-defined validators. I would love to get a second opinion from @stefanpenner @rwjblue or @blimmer before making any decisions.

elwayman02:
It seems like a pretty generic validator to me. What's your concern here, @offirgolan?

elwayman02:
A pretty common use-case for a validator like this would be a select that offers an "Other" option. Selecting any other option would be tied to a model with an id, but selecting "Other" would generally show a free-text input. On the parent model that has to validate this, you would need to either allow the id to be set for that field or have a separate attribute to contain the free-text input. You can't enforce a normal `presence` validator on either attribute as a result.

offirgolan:
Hi @elwayman02, after speaking with Stef, We both think this validator is great but a bit esoteric and therefore would be better implemented as a separate addon. If you're up for it, I can create a template repo for 3rd party validators to help you get started. Once done, I can add it to the readme to get some more visibility. This would be a great starting point for other people as well :smile:

elwayman02:
I'm not sure I agree that it's esoteric in nature. An exclusive-or is a very basic operation in mathematics and computer science, and it's a relatively common scenario for a validation as well, as explained above. Splitting things like this out into a series of "plugin"-like addons is a slippery slope for consumers; there are a lot of problems with that approach. However, I fully support keeping the core library lean and focused on the most common functionality. Can you or @stefanpenner help me understand why you think this validator is too specialized for the core addon?

elwayman02:
As an aside, I think having a template repo for 3rd-party validators is a great idea @offirgolan

PR #59, opened 2015-10-21
Add enumerable list of all error objects

davewasmer:
Currently, `validations.messages` exposes a flattened array of all error messages present. It would be useful (for a variety of reasons) to have access to both the message as well as the attribute name it corresponds to.

This PR adds `validations.errors` following the same pattern as `validations.messages` (it also adds `validations.error`).

(docs included)

### DIFF
--- a/addon/validations/error.js
+++ b/addon/validations/error.js
@@ -1,0 +1,6 @@
+import Ember from 'ember';
+
+export default Ember.Object.extend({
+  attribute: null,
+  message: null
+});
--- a/addon/validations/result.js
+++ b/addon/validations/result.js
@@ -5,6 +5,7 @@
 
 import Ember from 'ember';
 import ValidationResultCollection from './result-collection';
+import ValidationError from './error';
 import { hasEmberData } from '../utils/utils';
 
 const {
@@ -68,6 +69,17 @@ var ValidationsObject = Ember.Object.extend({
       }
     }
     return !isNone(attrValue);
+  }),
+
+  error: computed('message', 'isInvalid', 'attribute', function() {
+    if (get(this, 'isInvalid')) {
+      return ValidationError.create({
+        message: get(this, 'message'),
+        attribute: get(this, 'attribute')
+      });
+    } else {
+      return null;
+    }
   })
 });
 
@@ -84,6 +96,8 @@ export default Ember.Object.extend({
   isDirty: computed.oneWay('_validations.isDirty'),
   message: computed.oneWay('_validations.message'),
   messages: computed.oneWay('_validations.messages'),
+  error: computed.oneWay('_validations.error'),
+  errors: computed.oneWay('_validations.errors'),
 
   // This hold all the logic for the above CPs. We do this so we can easily switch this object out with a different validations object
   _validations: computed('model', 'attribute', '_promise', function() {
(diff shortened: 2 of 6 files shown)


rwjblue:
I think this will need some tests. Also, can you tweak the docs to indicate what the contents of .errors will be?

davewasmer:
Good idea - caught a mistake (although I have a sneaking suspicion you knew that already :wink: )

Let me know if you'd like me to squash, or if I missed something in my test coverage.

offirgolan:
I like this idea but I don't believe this is the best approach. If you look at how this is structured, all validation objects share the same structure. Meaning that the global props, validation-result, and validation-result-collection all have (messages, invalid, valid, etc.). All this is currently doing is putting `errors` on the global validation object.I would suggest to start from `validation-result.js` and work up.

davewasmer:
@offirgolan glad to hear you like the idea.

So adding an `error`/`errors` property to each level? I understand the structure you are describing, but I'm not sure how that would work, since the error objects basically *are* the validation-result objects. 

I could introduce a new concept (an "Error object" for lack of a better name) that is essentially just the message & attribute name, and attach those at each level. Is that what you had in mind?

offirgolan:
Since all validations are composed of validation `result-collection` objects, I feel like we can put the logic there (i dont think you need to touch the `result` file). So `errors` can listen to `messages.[]` and generate a set of error objects

```js
{ 
  attribute: get(this, 'attribute'), 
  message: message
}
```

Or like you said, we can introduce a new concept `addon/validations/error.js`

```js
export default Ember.Object.extend({
 attribute: null,
 message: null
})
```
And just create these. I kind of prefer the latter since it can scale better? 

From the global props perspective, it can be identical to how the global messages property is computed. 

@rwjblue thoughts?

offirgolan:
Actually now that I think of it, the initial error object creation should happen on the `result` object, not on the `result-collection` object. So the `error` prop will listen to `message` which will then create an error object with it's own attribute. On the `result-collection` object, it would just be the same logic as the messages and same for the global prop.

davewasmer:
@offirgolan alrighty, updated. Let me know how that looks.

addon/validations/error.js
```diff
@@ -0,0 +1,8 @@
+import Ember from 'ember';
+
+export default Ember.Object.extend({
+
+  attribute: null,
+  message: null
+
```
offirgolan: Can you remove these empty lines (before and after the props)

offirgolan:
This looks great! If you can just fix that one comment and rebase this to a single commit, I will merge this :smiley_cat:

addon/validations/result.js
```diff
@@ -68,6 +69,21 @@ var ValidationsObject = Ember.Object.extend({
...
+      return ValidationError.create({
+        message: get(this, 'message'),
+        attribute: get(this, 'attribute')
+      });
+    } else {
+      return null;
+    }
+  }),
+
+  errors: computed('error', function() {
```
offirgolan: This isnt needed here. A result will only have a single error unless the _validations object get replaced by a validation-result object (only in belongs-to and has-many). Thats why the messages CP exists here but no actually messages prop.

davewasmer:
@offirgolan address both comments and rebased!

offirgolan:
Beautiful! :100: thank you sir!

davewasmer:
:beers:

PR #61, opened 2015-10-26
Uniform result and result-collection

offirgolan:
This was pointed out by @davewasmer that result and result-collection
is not uniform. Doing this small change removes some unnecessary code.


stefanpenner:
GetEach relies on prototype extensions.

offirgolan:
True. Content is set appropriately ([see here](https://github.com/offirgolan/ember-cp-validations/pull/61/files#diff-7abb102ef4fa1b1eb69bbef7bc59a958R36))

stefanpenner:
Nice

Issue #60, opened 2015-10-23
Number validator?


dzbo:
I can't locate number validator neither in code nor in the docs. I see there is message for that https://github.com/offirgolan/ember-cp-validations/blob/master/addon/validators/messages.js#L78. Should I create custom one then?

offirgolan:
My goal was to be able to replace the number validator with the [inclusion/exclusion validators](http://offirgolan.github.io/ember-cp-validations/docs/validators/inclusion/index.html) since they are type checked. If this isnt enough, just use a custom validator for now. If there is a need for a number validator we can create one later down the line.

dzbo:
Yep, I've created my own number validator for now. But yea, it would be great to have one built in, it's rather common.

offirgolan:
Please upgrade to 2.1.0 and check out the [number validator docs](http://offirgolan.github.io/ember-cp-validations/docs/validators/number/index.html) :smile_cat:

PR #62, opened 2015-10-26
number validator

offirgolan:
I initially thought this can be replaced by type checked
inclusion/exclusion but there are way too many specific use cases.

This should also suffice #60


stefanpenner:
LGTM

PR #64, opened 2015-11-03
add allowBlank option to number validator

jakesjews:


app/validators/number.js
```diff
@@ -12,6 +12,10 @@ const {
 
 export default Base.extend({
   validate(value, options) {
+    if (options.allowBlank && isEmpty(value)) {
```
offirgolan: To be consistent with the rest of the validators, can you move this block below the variable declaration

offirgolan:
Ah I totally forgot about this when I wrote this one. Thanks for the PR :D. Besides my note, can you also add the allowBlank options in the docs (docs/docs/validators/number.md). You can just copy it from one of the other ones.

jakesjews:
No problem I'll get that in real quick

offirgolan:
Looks good. Thank you :smile_cat:

Issue #63, opened 2015-10-30
Presence validation fails on relationships


elwayman02:
If you have a `belongsTo` or `hasMany` and simply want to check that _something_ has been set on the record, but the models on the other side of the relationship don't have validations to run, it makes sense to use a `presence` validator on that attribute. Unfortunately, if you're using Ember-Data (for example), the relationship actually creates a "fulfilled" model that has `null` content until an actual record is set on that relationship. The `presence` validator naively checks and sees that there is an object there, and returns true to indicate it's vaild, when really it is not. Thoughts?

elwayman02:
To clarify, Ember-Data creates a `DS.PromiseObject` on the belongsTo relationship, whose content is generally the assigned `DS.Model` record, but is null in the case of nothing being assigned?

elwayman02:
I believe the same issue will exist for a `hasMany`, where a `DS.PromiseArray` is initialized by default

elwayman02:
Something like this may work:
```
keyPresent(model, key) {
    const value = model.get(key);
    if (isBlank(value)) {
      return false;
    }
    const constructor = value.constructor;
    if (isEqual(typeOf(value.get), 'function') && isPresent(constructor) && isEqual(typeOf(constructor.toString), 'function')) {
      const type = constructor.toString();
      if (['DS.PromiseObject', 'DS.PromiseArray'].indexOf(type) !== -1) {
        return isPresent(value.get('content'));
      }
    }
    return true;
  },
```
Curious as to your thoughts, @offirgolan

elwayman02:
Never mind, I've confirmed that this only occurs for `belongsTo`, not `hasMany`. With `hasMany`, Ember-Data leaves the attribute as null until something is set, it seems.

elwayman02:
Related: https://github.com/emberjs/data/issues/3897

PR #65, opened 2015-11-03
Handle ember-data promise proxy based instances in presence

offirgolan:
Regarding #63

### DIFF
--- a/app/validators/presence.js
+++ b/app/validators/presence.js
@@ -7,6 +7,7 @@ import Ember from 'ember';
 import Base from 'ember-cp-validations/validators/base';
 
 const {
+  get,
   isEmpty
 } = Ember;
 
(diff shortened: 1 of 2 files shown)


offirgolan:
This is ready for review but Im not really sure if it's the best way to handle this. Suggestions are definitely welcome.

app/validators/presence.js
```diff
@@ -21,14 +23,27 @@ export default Base.extend({
...
     return true;
+  },
+
+  /**
+   * Handle presence of ember-data promise proxy based instances
+   */
+  _isPresent(value) {
+    if(typeOf(value) === 'instance' && value.constructor) {
+      const type = value.constructor.toString();
+      if(['DS.PromiseObject', 'DS.PromiseArray'].indexOf(type) !== -1) {
```
jasonmit: ArrayProxy and ObjectProxy would handle both cases since DS.PromiseObject and DS.PromiseArray extend from them.  This would cover the case of comparing an object which is of those 4 types.

As for comparing, I'd lean towards using `detect` which will walk the inheritance graph.

`Ember.ArrayProxy.detect(foo); => true/false`

stefanpenner: :+1:

offirgolan:
Using `detectInstance` seemed to get the job done :smile:

offirgolan:
@stefanpenner good to merge?

app/validators/presence.js
```diff
@@ -21,14 +22,24 @@ export default Base.extend({
...
     }
 
     return true;
+  },
+
+  /**
+   * Handle presence of ember proxy based instances
+   */
+  _isPresent(value) {
+    if(Ember.ObjectProxy.detectInstance(value) || Ember.ArrayProxy.detectInstance(value)) {
```
stefanpenner: i would recommend using the raw JS `instanceof` because this is what `detectInstance` now becomes
https://github.com/emberjs/ember.js/blob/d38cadd99aa1fbeea5b603b4eff2b94fb3fe74ec/packages/ember-runtime/lib/system/core_object.js#L811-L813

We plan to likely remove detectInstance in 3.0

jasonmit: Good to know, thanks

offirgolan:
alright made the final changes. Feel free to merge.

PR #70, opened 2015-11-09
Update README.md

kellyselden:
I think you want this to be the same as the below example.


offirgolan:
Good catch. Thank you :smile_cat:

PR #76, opened 2015-11-19
Makes isDirty work properly for boolean

mjuniper:
For boolean attributes with defaultValue === false, isDirty always returned true. This should fix that. All tests pass.

### DIFF
--- a/addon/validations/result.js
+++ b/addon/validations/result.js
@@ -64,7 +64,7 @@ var ValidationsObject = Ember.Object.extend({
       let attrMeta = model.get('constructor.attributes').get(attribute);
       if (attrMeta) {
         let defaultValue = attrMeta.options.defaultValue;
-        if (defaultValue) {
+        if (!isNone(defaultValue)) {
           return defaultValue !== attrValue;
         }
       }
--- a/tests/dummy/app/models/signup.js
+++ b/tests/dummy/app/models/signup.js
@@ -1,0 +1,19 @@
+import DS from 'ember-data';
+import {
+  validator, buildValidations
+}
+from 'ember-cp-validations';
+
+
+let Validations = buildValidations({
+  name: validator('presence', true),
+  acceptTerms: validator((value) => {
+    return value || 'You must accept the terms.';
+  }),
+});
+
+
+export default DS.Model.extend(Validations, {
+  name: DS.attr('string', { defaultValue: '' }),
+  acceptTerms: DS.attr('boolean', { defaultValue: false }),
+});
--- a/tests/unit/validations/ds-model-test.js
+++ b/tests/unit/validations/ds-model-test.js
@@ -1,0 +1,34 @@
+import Ember from 'ember';
+import {
+  moduleForModel,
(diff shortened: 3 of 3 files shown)


rwjblue:
Looks good, would you mind adding a test that fails on master and is fixed by this tweak?

addon/validations/result.js
```diff
@@ -64,7 +64,7 @@ var ValidationsObject = Ember.Object.extend({
       let attrMeta = model.get('constructor.attributes').get(attribute);
       if (attrMeta) {
         let defaultValue = attrMeta.options.defaultValue;
-        if (defaultValue) {
+        if (!Ember.isNone(defaultValue)) {
```
offirgolan: Can you change this from `!Ember.isNone` to `!isNone` since its already defined as a const up top

mjuniper:
Sure, but maybe you can point me in the right direction. I'd like to just define a model on the fly in my test and then create it. Obviously I can't just do `model.create`, I need to do `store.createRecord`. So I need to have access to the store....

I've done this:

```
module('Unit | Validations | DS.Model', {
  beforeEach() {
    registry = new Registry();
    container = registry.get('container');

    container.register('store:main', DS.Store.extend());
    container.register('model:signup', Model);
    store = container.lookup('store:main');
  }
});
```

But when I do `store.createRecord('signup')`, I get `Assertion Failed: You need to provide an object and key to `set`.` Any suggestions? Maybe I'm just taking a completely wrong approach?

mjuniper:
Ok, I got it working by just creating a model in the dummy app and using moduleForModel in my test. Please let me know if there is a better way.

mjuniper:
I'm not sure why the build is busted - it looks like it never even gets to the tests....

tests/unit/validations/ds-model-test.js
```diff
@@ -0,0 +1,36 @@
+import Ember from 'ember';
+import DS from 'ember-data';
+import Registry from '../../helpers/registry';
```
offirgolan: this import is making the tests fail

mjuniper:
Weird, my linter usually yells at me for unused imports. Anyhow, it is passing now.

offirgolan:
Awesome :+1: ! Can you just do a quick rebase to reduce the number of commits? Ill merge after

mjuniper:
:+1:

tests/dummy/app/models/signup.js
```diff
@@ -0,0 +1,19 @@
+import DS from 'ember-data';
+import {
+  validator, buildValidations
+}
+from 'ember-cp-validations';
+
+
+let Validations = buildValidations({
+  name: validator('presence', true),
+  acceptTerms: validator((value) => {
```
offirgolan: Just as a note, you can use the inclusion validator here

```js
validator('inclusion', {
  in: [true],
  message: 'You must accept the terms'
})
```

offirgolan:
Thanks for the PR @mjuniper :smile_cat:

Issue #81, opened 2015-11-20
Clear error messages


offirgolan:
In reference to #80 

There should be a way to clear errors with the ability to clear errors for a specific type(s).

- [x] `clearErrors()`
- [x] `clearErrors(['presence', 'length'])`

offirgolan:
I was able to approach this in a much simpler way that doesnt require any change for this library. I created a demo app to illustrate this

https://github.com/offirgolan/clear-cp-demo

Feel free to clone, `npm i && bower i` then `ember s` to see it in action. The demo is based on the demo app of this library but with a reset functionality. To see the changes I made, you can look at [this commit](https://github.com/offirgolan/clear-cp-demo/commit/0b22306da87396c0a7388150862c04b182ae3d40)

frank06:
I can't seem to find any connection between the "reset functionality" in the sample app (which overrides a component and is pretty contrived using `didValidate`) and why `clearErrors` or similar have not been implemented.

In a component I have a property called `current` which backs a form. It is a fragment, created via `this.store.createFragment`. When `current` is saved, it's sent up via an action. Until here validations work perfectly well. Then, `current` is then reset to a brand-new fragment.  When that happens all blank fields now show validation errors (presence).

I need a way to tell `ember-cp-validations` that the model/fragment is a new, untouched one. That it should not attempt to validate until the user tries to make a selection/change. So, how can I do that? Thanks!

escobera:
I'm needing this as well. @offirgolan What is the current way to achieve this behaviour?

PR #82, opened 2015-11-20
Clear errors implementation

offirgolan:
Solves #81 

I do have a couple comments/concerns: 

1. Is clearErrors the correct naming for this?
2. When we clear an error message, should the result object be set to `isValid: true`? Or should we just clear the message?
3. Instead of updating the result object, maybe removing it all together from the result collection would make more sense?

Usage:

```js
obj.clearErrors();
obj.clearErrors('presence');
obj.clearErrors(['presence', 'length']);

obj.get('validations.attrs.foo').clearErrors();
// ....

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,8 @@ env:
   - EMBER_TRY_SCENARIO=ember-1.11
   - EMBER_TRY_SCENARIO=ember-1.12
   - EMBER_TRY_SCENARIO=ember-1.13
+  - EMBER_TRY_SCENARIO=ember-2.0
+  - EMBER_TRY_SCENARIO=ember-2.1
   - EMBER_TRY_SCENARIO=ember-release
   - EMBER_TRY_SCENARIO=ember-beta
   - EMBER_TRY_SCENARIO=ember-canary
--- a/addon/validations/factory.js
+++ b/addon/validations/factory.js
@@ -60,7 +60,8 @@ export default function buildValidations(validations = {}) {
   var GlobalValidations = Ember.Object.extend(props, {
     isValidations: true,
     validate,
-    validateSync
+    validateSync,
+    clearErrors
   });
 
   return createMixin(GlobalValidations, AttrValidations);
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by rox163** [comment]
IMHO -
For the naming, maybe clearValidationErrors? I guess clearErrors is a bit vague.
When we clear, I think it should be 'reset' to the state it was in when we first visited the page. I think isValid is set to false at that point, because the field would have no content? If isValid is true, then that might cause issues with styling the field based on its validity or maybe even submit button styling based on validity.

**Comment by offirgolan** [comment]
After some more testing, I've found that this problem is more complicated than I expected. The issue lies with modifying a result object outside of the main CP. If we clear a validation to its default state ( `message: null, isValid:true` ) the validation CP will not be updated again until its dependent keys change.

1. validate --> isValid: false
2. clearErrors --> isValid:true
3. validate --> isValid:true. This should ideally rerun the validations but really, nothing actually changed

Im sort of leaning to not needing to support this use case since it is more of a UX issue. Instead, maybe propogate validation types.

If you want to display all errors that are not `presence`, you can instead just do something like this...

```js
errors: computed('validations.errors', function() {
 return this.get('validations.errors').filter(e => e.type !== 'presence')
})
```

```hbs
{{#each errors as |error|}}
 {{error.message}}
{{/each}}
```

rox163:
Well, in my case I was just trying to reset the form completely, clearing all errors and all fields, not specifics fields/errors. So I thought this could be driven by the library. I see your point about the specific error message display being a UX issue.

offirgolan:
Gotcha. Let me spend more time on this and see if I can come up with a cleaner and better working solution.

rox163:
So far this is what I added to my form component to try to clear error messages and content from all fields when Clear button is clicked - 
```
clearForm() {
this.get('fieldIds').forEach( (fieldId) => {
		this.set('form_' + fieldId, '');
		this.set('didValidate', false);
		this.set('validations.attrs.' + fieldId + '.isInvalid', false);			
                this.set('validations.attrs.' + fieldId + '.message', null);
		$('#form_' + fieldId).val("");
	});
}
```
Seems to work. I do see intermittent issues where sometimes I need to click on the button twice for it to work.

rox163:
Got back to fixing my form up this week. Setting the invalid property and message doesnt work properly in my component, when I try to Clear the form of all errors and values. 
```
this.set('validations.attrs.' + fieldId + '.isInvalid', false);         
this.set('validations.attrs.' + fieldId + '.message', null);
```
First it says I must use Ember.set() to set these values. Then when i use Ember.set(), it says I cannot set these values because they are readonly.

offirgolan:
That wont work since attrs.fieldId is a CP made of other CPs

You will have to do the following

```js
this.get(`validations.attrs.${fieldId}.content`).forEach(v => v.update({isValid: true, message: null}))
```

rox163:
Thanks! That helped!
Next hurdle in this workflow - So clearing the form now works and the field values are cleared in the model too. The problem with setting isValid:true on each field though is if I clear the form and then click  my Submit button, it submits the empty model because it thinks all the fields are valid :(
```
 validateAndSubmit() {
      let {model, validations} = this.validateSync();
      this.set('didValidate', true);
      if (validations.get('isValid')) {
        this.get('on-submit')(model);
      }
    }
```

offirgolan:
@rox163 I was able to approach this in a much simpler way that doesnt require any change for this library. I created a demo app to illustrate this

https://github.com/offirgolan/clear-cp-demo

Feel free to clone, `npm i && bower i` then `ember s` to see it in action. The demo is based on the demo app of this library but with a reset functionality. To see the changes I made, you can look at [this commit](https://github.com/offirgolan/clear-cp-demo/commit/0b22306da87396c0a7388150862c04b182ae3d40)

Issue #86, opened 2015-12-10
validators of statically created objects can't access custom Messages object  


ramybenaroya:
In validators of statically created objects `this.errorMessages` points to the default `ember-cp-validations/validators/messages` instead of `app/validators/messages`

For example:
```javascript
// app/validators/messages.js
export default Messages.extend({
  example: "example error message"
});
```

```javascript
// models/user.js
var Validations = buildValidations({
  username: [
     validator(function(){
        return this.errorMessages.getMessageFor('example', {});
     })
  ]
});
export default Ember.Object.extend(Validations, {
  username: null
});
```

```javascript
// routes/index.js
import User from '../models/user';
export default Ember.Route.extend({
  model() {
    var options = { username: 'John' };
    Ember.setOwner(options, Ember.getOwner(this));
    return User.create(options)
  }
});
```

```hbs
{{!-- /app/templates/index.hbs --}}
{{v-get model 'username' 'message'}}
```

The output will be `"undefined is invalid"` instead of `"example error message"`.   
This is because the default message type is `invalid`.

offirgolan:
Ah I know what the issue is. Ill open a PR with the fix in a little bit. Thanks for pointing this issue out!

offirgolan:
Waiting on rwjblue/ember-getowner-polyfill#1 before this can be resolved.

offirgolan:
@ramybenaroya please upgrade to `v2.1.4`. Should solve this problem :smile_cat:

ramybenaroya:
@offirgolan Thanks for all your hard work and for this great addon!

offirgolan:
Np :smiley_cat: Feel free to reach out if any other issues come up!

PR #90, opened 2015-12-16
Fix owner problems in 2.2.0+

offirgolan:
In reference to #86


offirgolan:
@rwjblue @stefanpenner can I get a quick review on this. This is a temporary fix until a setOwner polyfill is created. Also, ignore the build failing, its due to components/ember-data#34

app/validators/date.js
```diff
@@ -5,7 +5,7 @@
 
 import Ember from 'ember';
 import Base from 'ember-cp-validations/validators/base';
-const moment = self.requirejs.entries['moment'] && self.require('moment')['default'];
+const moment = (self.requirejs.entries['moment'] || self.requirejs.entries['moment/index']) && self.require('moment')['default'];
```
rwjblue: We should add a nice `loaderjs.has` helper method for this, but that is another issue altogether :smiley_cat:

rwjblue:
@offirgolan - This looks like a good workaround for now, sorry I've been so delayed in adding support upstream.

app/validators/date.js
offirgolan: That would definitely be nice :smile:

offirgolan:
@rwjblue no rush! This should hold down the fort for a while

Issue #58, opened 2015-10-20
Debounced Validations


offirgolan:
In regards to #57, it might be useful from a UX perspective to have an option to debounce validations. This would also be needed for custom ajax dependent validations. This will definitely need to be explored a little more but a simple definition can look something like this: 

```javascript
var Validations = buildValidations({
  foot: validator('custom-ajax', {
    debounce: 500
  })
});
```

@blimmer @stefanpenner any thoughts?

jamesdixon:
This would be great. :+1:

stefanpenner:
is it debouncing we want or throttle?

stefanpenner:
It seems like we actually want tighter integration with various events, focusOut etc. Having a bunch of async validations on a form the user can fill out quickly results in this strange laziness. 

I suspect some combination of better event hooks, and debounce is what we would like to see. (maybe this doesn't preclude standalone debounce)

ajgordo:
I'm curious if there is any kind of workaround for forcing validation on focusOut right now. Using the validated-input component as an reference I want to add to it a validate check when it loses focus. I see you can do that on the model, but I want to see if you can do it within the component itself?

stefanpenner:
> I'm curious if there is any kind of workaround for forcing validation on focusOut right now.

I would love for this to be easy. I think @offirgolan and me have chatted about it a few times.

jamesdixon:
@ajgordo not sure if this helps as I'm using [ember-paper](https://github.com/miguelcobain/ember-paper) for my component library, but I modified `validated-input` like so to get validation working on focusOut:

```javascript
// validated-input.js
import Ember from 'ember';

const {
  computed,
  defineProperty,
  run
} = Ember;

export default Ember.Component.extend({
  model: null,
  value: null,
  rawInputValue: null,
  type: 'text',
  textarea: false,
  valuePath: '',
  placeholder: '',
  label: '',
  attributeValidation: null,
  showError: false,

  isValid: computed.oneWay('attributeValidation.isValid'),

  isInvalid: computed.oneWay('attributeValidation.isInvalid'),

  didInitAttrs() {
    this._super(...arguments);
    const valuePath = this.get('valuePath');
    defineProperty(this, 'attributeValidation', computed.oneWay(`model.validations.attrs.${valuePath}`));
    defineProperty(this, 'value', computed.alias(`model.${valuePath}`));
  },

  actions: {

    validate() {
      this.set('showError', this.get('isInvalid'));
    }
[...]

Issue #57, opened 2015-10-20
[Question] Using a debounce for a promise-based validation


dasnixon:
I have a validation that makes a `GET` request to my server to check if a resource already exists in the system. What is the best approach for a `debounce`? I don't want to make a zillion requests to my API. I tried a few things to no avail. Maybe the debounce logic should just live outside the scope of the validation? To me it feels like I should handle this where the actual request occurs. Hopefully someone can point me in the right direction, thanks!

````
import Ember from 'ember';
import BaseValidator from 'ember-cp-validations/validators/base';

const { isPresent } = Ember;

export default BaseValidator.extend({
  validate(value, options, _model, attribute) {
    const adapter = this.container.lookup('adapter:unique-resource');
    return adapter.checkUniqueness(attribute, value, options.endpoint).then(function () {
      return true;
    }).catch(() => {
      const errorMessage = isPresent(options.message) ? options.message : this._defaultMessage(attribute);
      return this.createErrorMessage('invalid', value, { message: errorMessage });
    });
  },

  _defaultMessage(attribute) {
    return `${attribute} has already been taken`;
  }
});
````

**Comment by offirgolan** [comment]
If your attribute is bound to an input I would put my debounce on the input value binding. 

Lets say my validated attribute is `foo`, then instead of binding my input to `foo`, you can bind to temporary property and debounce that. 

```handlebars
{{input value=rawFooValue}}
```

```javascript
export default Ember.Component.extend({
  rawFooValue: null,

  onFooChange: Ember.observer('rawFooValue', function() {
    Ember.run.debounce(this, this._setFoo, 500, false);
  }),

  _setFoo() {
    this.set('foo', this.get('rawFooValue')); // This will now trigger your validation rules
  }
})
```

Hope that helps! You can take a look [here](https://github.com/offirgolan/ember-cp-validations/blob/master/tests/dummy/app/components/validated-input.js#L39) for a more detailed example

dasnixon:
Makes sense, thanks for the feedback. Thought maybe there was some more elegant solution :smile_cat:

offirgolan:
Hmmm it might be useful to have a debounce option in the validator options. Ill explore this more :smile:

dasnixon:
Sweet, I might play around a little, I'll let you know what I find. I'll re-open this since it _could_ be useful

dasnixon:
@offirgolan thanks for implementing!

offirgolan:
[...]

PR #92, opened 2015-12-25
Debounced validation functionality

offirgolan:
In regards to #57 & #58 

```javascript
var Validations = buildValidations({
  foo: validator('custom-ajax', {
    debounce: 500
  })
});
```

@stefanpenner @rwjblue let me know what you guys think of this.


tests/integration/validations/factory-general-test.js
```diff
@@ -247,3 +247,35 @@ test("custom messages object", function(assert) {
...
+  var done = assert.async();
+  var Validations = buildValidations({
+    firstName: validator(Validators.presence),
+    lastName: validator(Validators.presence, { debounce: 500 }),
+  });
+  var object = setupObject(this, Ember.Object.extend(Validations));
+
+  assert.equal(object.get('validations.isValid'), false, 'isValid was expected to be FALSE');
+  // TODO: I feel like initially a debounced validation should not be debounced.
+  assert.equal(object.get('validations.isValidating'), true, 'isValidating was expected to be TRUE');
```
offirgolan: Right when we create the object, the `lastName` debounced promised is created which sets the `isValidating` to true

tests/integration/validations/factory-general-test.js
```diff
@@ -247,3 +247,35 @@ test("custom messages object", function(assert) {
...
+  assert.equal(object.get('validations.isTruelyValid'), false, 'isTruelyValid was expected to be FALSE');
+
+  Ember.run.later(() => {
```
offirgolan: I dont know if this is the correct behavior. Instantly, we have to wait for all the debounced validations to finish before we can do anything else. I guess one solution is just to call `object.validate().then(...)` since I considered the debounced validations as async

tests/integration/validations/factory-general-test.js
```diff
@@ -247,3 +247,35 @@ test("custom messages object", function(assert) {
...
+    assert.equal(object.get('validations.attrs.lastName.isValid'), false);
+    assert.equal(object.get('validations.attrs.lastName.isValidating'), false);
+    assert.equal(object.get('validations.attrs.lastName.message'), 'lastName should be present');
+
+    object.set('lastName', 'Golan');
+    // TODO: This is true because a new Validation result object is created which initially defaults to TRUE
+    assert.equal(object.get('validations.attrs.lastName.isValid'), true);
```
offirgolan: Again, this doesnt feel right. When we set a new value to the debounced CP, it creates a new ValidationResult object with the new promise but that new object gets instantiated by default with `isValid: true`

stefanpenner:
does anything cancel them, if the object is destroyed?

offirgolan:
Ah. No there is not cleanup. Let me do that :smile:

offirgolan:
@stefanpenner all done. Let me know if you approve :dancers:

stefanpenner:
@offirgolan ping me around noonish tomorrow, i gotta clear my queue of other stuff before i do a thorough review.

offirgolan:
Will do!

addon/validations/factory.js
```diff
@@ -205,28 +208,29 @@ function createCPValidationFor(attribute, validations) {
...
 
-    validationResults = flatten(validationResults);
-    var validationResultsCollection = ValidationResultCollection.create({
-      attribute, content: validationResults
-    });
+      if(debounce > 0) {
+        let cache = getDebouncedValidationsCacheFor(model);
+        // Return a promise and pass the resolve method to the debounce handler
+        value = new Promise(resolve => {
+          cache[attribute] = run.debounce(validator, getValidationResult, validator, options, model, attribute, resolve, debounce, false);
```
stefanpenner: We likely need to deal with the > 1 debounced per attribute, and also > 1 debounced per attribute per type (2 ajax validators for 1 attribute example).

addon/validations/factory.js
```diff
@@ -272,17 +276,36 @@ function getCPDependentKeysFor(attribute, validations) {
...
+ * Used to retrieve the validation result by calling the validate method on the validator.
+ * If resolve is passed, that means that this validation has been debounced so we pass the
+ * result to the resolve method.
+ * @param  {Validator} validator
+ * @param  {Object} options
+ * @param  {Object} model
+ * @param  {String} attribute
+ * @param  {Function} resolve
+ */
+function getValidationResult(validator, options, model, attribute, resolve) {
```
stefanpenner: instead of this function, we should likely just do the following, from the calling side.

```
_ => resolve(validator.validate(get(model, attribute), options, model));
```

addon/validations/factory.js
```diff
@@ -445,3 +484,21 @@ function validate(options = {}, async = true) {
...
   return this.validate(options, false);
 }
+
+function willDestroy() {
+  this._super(...arguments);
+  const caches = ['_validators', '_debouncedValidations'];
+  const key = getKey(this);
+
+  // Cancel all debounced timers
+  let debounceCache = get(this, `validations._debouncedValidations.${key}`) || {};
```
stefanpenner: maybe a more OO style encapsulation, something like `get(this, 'validations._debouncedValidations').destroy();`

offirgolan:
@stefanpenner seems like travis keeps failing because it can find `Map` since I started using `new Map()` for my cache... do you have a workaround on this? Or should I just scrap using Map?

Also, ignore all the files.... something got messed up when I tried to merge so just look at the commit

offirgolan:
I decided to revert back to just using pojos as the caches. I really dont think there is much of a need to create another class for such a simple use case. In the future, if it becomes needed, ill make the refactor and hopefully by then es6 Map will be more widely supported (phantomjs included :disappointed: ). 

As for the `willDestroy`, again, I dont think it's very complicated to need even more deconstruction... if you're still adamant on these let me know and we can chat :smile:

PR #96, opened 2016-01-07
validation result options hash

### LINKED ISSUE #93: Implement public API for getting options from each validation.
HI there! Great work you've been doing. Thanks!

I haven't found a public API for getting the original options passed to each validation. I would like to use them in association with certain HTML attributes. Here is an example of what I would like to do:

``` handlebars
{{input value=model.email
        required=model.validations.attrs.email.validations.presence
        maxlength=model.validations.attrs.email.validations.length.max}}
```

I am using the ```v-get``` helper. I wrote the example that way trying to make myself clearer.

Anyway, I think that such API would be useful.

offirgolan:
Should resolve #93 

All built options of the validators associated with the results in this collection grouped by validator type

```javascript
// Given the following validators
{
  username: [
    validator('presence', true),
    validator('length', { max: 15 }),
    validator('format', { regex: /foo/ }),
    validator('format', { regex: /bar/ }),
  ]
}
```

```js
get(user, 'validations.attrs.username.options')
```

The above will return the following
```js
{
  'presence': { presence: true},
  'length': { max: 15 },
  'regex': [{ regex: /foo/ }, { regex: /bar/ }]
}
```

This will also allow you to do the following: 

```hbs
{{input value=model.username
        required=model.validations.attrs.username.options.presence
        maxlength=model.validations.attrs.username.options.length.max}}
```

### DIFF
--- a/addon/validations/factory.js
+++ b/addon/validations/factory.js
@@ -253,7 +253,7 @@ function createCPValidationFor(attribute, validations) {
         value = validator.validate(attrValue, options, model, attribute);
       }
 
-      return validationReturnValueHandler(attribute, value, model);
+      return validationReturnValueHandler(attribute, value, model, validator);
     });
 
     return ValidationResultCollection.create({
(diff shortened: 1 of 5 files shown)


stefanpenner:
:+1:

PR #98, opened 2016-01-08
Use ember-cli-release to automate dummy app deploy

offirgolan:
Use ember-cli-release and some random fixes

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,23 +1,23 @@
 {
   "name": "ember-cp-validations",
   "dependencies": {
-    "ember": "1.13.8",
+    "ember": "1.13.10",
     "ember-cli-shims": "0.0.6",
     "ember-cli-test-loader": "0.2.2",
     "ember-data": "~2.0.0",
     "ember-load-initializers": "0.1.7",
     "ember-qunit-notifications": "0.1.0",
     "ember-resolver": "~0.1.20",
-    "jquery": "^1.11.1",
+    "jquery": "1.11.3",
     "loader.js": "ember-cli/loader.js#3.2.1",
     "qunit": "~1.20.0",
     "moment": ">= 2.8.0",
     "moment-timezone": ">= 0.1.0"
   },
   "devDependencies": {
-    "blanket": "~1.1.5"
+    "blanket": "~1.1.7"
   },
   "resolutions": {
-    "ember": "1.13.8"
+    "ember": "1.13.10"
   }
 }
--- a/config/release.js
+++ b/config/release.js
@@ -1,0 +1,18 @@
+/* jshint node:true */
+/* var RSVP = require('rsvp'); */
+var execSync = require('child_process').execSync;
+
+module.exports = {
+  publish: true,
+  afterPublish: function(project, versions) {
+    // Publish dummy app with docs to gh-pages
+    runCommand('ember github-pages:commit --message "Released ' + versions.next + '"');
+    runCommand('git push origin gh-pages:gh-pages');
+  }
+};
+
+function runCommand(command) {
+  console.log('running: ' + command);
+  var output = execSync(command, { encoding: 'utf8' });
+  console.log(output);
+}
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "ember-cli-inject-live-reload": "^1.3.0",
     "ember-cli-moment-shim": "0.7.1",
     "ember-cli-qunit": "^1.1.0",
+    "ember-cli-release": "^1.0.0-beta.0",
     "ember-cli-sri": "^1.2.0",
     "ember-cli-uglify": "^1.0.1",
     "ember-cli-yuidoc": "^0.8.2",
--- a/tests/dummy/app/index.html
+++ b/tests/dummy/app/index.html
@@ -18,7 +18,7 @@
 
     {{content-for 'head-footer'}}
   </head>
-  <body onload="prettyPrint()">
+  <body>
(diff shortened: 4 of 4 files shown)


addon/validations/factory.js
```diff
@@ -247,7 +247,7 @@ function createCPValidationFor(attribute, validations) {
         let cache = getDebouncedValidationsCacheFor(attribute, model);
         // Return a promise and pass the resolve method to the debounce handler
         value = new Promise(resolve => {
-          cache[getKey(validator)] = run.debounce(validator, () => resolve(validator.validate(attrValue, options, model, attribute)), debounce, false);
+          cache[getKey(validator)] = run.debounce(validator, () => resolve(validator.validate(attrValue, validator.processOptions(), model, attribute)), debounce, false);
```
offirgolan: There could be times where function based options are time sensitive
and need to be computed right before we call validate. Unfortunately this means having to call `processOptions` twice

config/release.js
```diff
@@ -0,0 +1,27 @@
+/* jshint node:true */
+var RSVP = require('rsvp');
+var exec = require('child_process').exec;
+
+module.exports = {
+  publish: true,
+  afterPublish: function(project, versions) {
+    // Publish dummy app with docs to gh-pages
+    return runCommand('ember github-pages:commit --message "Released ' + versions.next + '"').then(function() {
```
rwjblue: Since you control the system you are running on, I would suggest using `execSync` instead of `exec` here...

```js
var execSync = require('child_process').execSync;

function runCommand(command) {
  console.log('running: ' + command);
  var output = execSync(command, { encoding: 'utf8' });
  console.log(output);
}

module.exports = {
  publish: true,
  afterPublish: function(project, versions) {
    runCommand('ember github-pages:commit --message "Released ' + versions.next + '"');
    runCommand('git push origin gh-pages:gh-pages');
  }
};
```

package.json
```diff
@@ -49,6 +49,7 @@
     "ember-cli-inject-live-reload": "^1.3.0",
     "ember-cli-moment-shim": "0.7.1",
     "ember-cli-qunit": "^1.1.0",
+    "ember-cli-release": "0.2.8",
```
rwjblue: The `publish: true` option was added in https://github.com/lytics/ember-cli-release/pull/39 and is only available in 1.0.0-beta.0.  You may want to update the version here to the beta or call `runCommand('npm publish')` in the `config/release.js`...

addon/validations/factory.js
rwjblue: Not to quibble, but this change seems unrelated to the PR at hand to me.  Am I missing something?

offirgolan: Hehe yeah my bad, i thought about this super late last night and was too lazy to create a new PR

offirgolan: Let me move this to a separate PR

offirgolan:
@rwjblue fixed :smile_cat:

rwjblue:
:+1: LGTM

addon/validations/factory.js
stefanpenner: :+1:

config/release.js
```diff
@@ -0,0 +1,18 @@
+/* jshint node:true */
+/* var RSVP = require('rsvp'); */
+var execSync = require('child_process').execSync;
+
+module.exports = {
+  publish: true,
+  afterPublish: function(project, versions) {
+    // Publish dummy app with docs to gh-pages
+    runCommand('ember github-pages:commit --message "Released ' + versions.next + '"');
```
stefanpenner: these are likely async, i suspect you actually want them to run in series not concurrently.

rwjblue: I'm not sure I follow what you mean. These are currently being run in order (not at the same time). See the definition of runCommand below (it's using execSync). Have I misunderstood how execSync works?

stefanpenner:
An. So this requires a native extension to run on node 10?

rwjblue:
@stefanpenner - Yes, but this is only for publishing this addon itself and not for consumers.

rwjblue:
I would be fairly surprised if @offirgolan or yourself are running node 0.10 as your primary node environment for working on addons....

stefanpenner:
@rwjblue well our employer does ;)

bower.json
```diff
@@ -1,21 +1,21 @@
...
-    "ember": "1.13.8",
+    "ember": "1.13.10",
     "ember-cli-shims": "0.0.6",
     "ember-cli-test-loader": "0.2.2",
     "ember-data": "~2.0.0",
     "ember-load-initializers": "0.1.7",
     "ember-qunit-notifications": "0.1.0",
     "ember-resolver": "~0.1.20",
-    "jquery": "^1.11.1",
+    "jquery": "~1.11.3",
```
rwjblue: Pin to specific version for now. Releases of Ember will be out soon hopefully with the fix (it's already been fixed on master).

offirgolan:
I myself am running node 12 but I have no issue in just wrapping `exec` in a promise.

offirgolan:
@rwjblue I like the idea of using execSync but as a precautionary measure, I think it would be better to just use exec and wrap it in a promise.

stefanpenner:
> I myself am running node 12 but I have no issue in just wrapping exec in a promise.

if its for dev, what exists currently is totally fine :)

offirgolan:
@stefanpenner you guys are killin me here

stefanpenner:
> But if its just for dev, thats fine.

?

PR #99, opened 2016-01-09
Reprocess options after debounce

offirgolan:
There could be times where function based options are time sensitive and need to be computed right before we call validate. Unfortunately this means having to call `processOptions` twice.


offirgolan:
@stefanpenner any thoughts on this? 

The first requirement for processing the options is to get the debounce value which can be a function, the second time is to make sure that options are processes right before validate is called.

PR #103, opened 2016-01-17
Debounce should not use an anonymous function

### LINKED ISSUE #102: debounced validations does not work
The debounce is not really debouncing because of the following code: https://github.com/offirgolan/ember-cp-validations/blob/master/addon/validations/factory.js#L250

You cannot pass an anonymous function as the debounce method, but it should rather pass in a reference of a function (http://emberjs.com/api/classes/Ember.run.html#method_debounce).

offirgolan:
Resolves #102


offirgolan:
@stefanpenner this fix is pretty minor, so im just gonna go ahead and :shipit:

PR #106, opened 2016-01-21
Expose disabled validator option

### LINKED ISSUE #104: validation depending on status of other field
I'm trying to implement form validation via a component, which receives it's model from the route.
This is working fine for most fields, but there are 2 fields that only need to be validated when one of the other fields, a checkbox, is set to true.
How can I do this?

I tried via an action which is triggered whenever the checkbox is toggled, but I don't see how I can access the validate() or validateSync() method from within that action.

This is the code in route.js:
```javascript
import Ember from 'ember';

var registration = {
    company: false,
    companyName: '',
    vat: '',
    username: '',
    email: ''
};

export default Ember.Route.extend({
  ,model: function() {
    return registration;
  }
});
```

And this is the code in component.js:

```javascript
import Ember from 'ember';

var validations = buildValidations({
    companyName: [
        validator(function(value, options, model/*, attribute*/) {
            // In case of company account, the company name should not be empty
            if (model.company) {
                return (value);
            }
            else {
                return true;
            }
        }, {
            message: 'company name required'
        })
    ],
    vat: [
        validator(function(value, options, model/*, attribute*/) {
            // In case of company account, the VAT number should not be empty
            if (model.company) {
                return (value);
            }
            else {
                return true;
            }
        }, {
            message: 'vat number required'
        })
    ],
    username: [
        validator('presence', {
            presence: true,
            message: ' '
        }),
        validator('length', {
            min: 3,
            message: ' '
        })
    ],
    email: [
        validator('presence', {
            presence: true,
            message: ' '
        }),
        validator('format', {
            type: 'email',
            message: 'invalid email address'
        })
    ]
});

export default Ember.Component.extend(validations, {
    registration: function() {
        return this.get('model');
    },
    actions: {
        // Explicit validation when checkbox is toggled:
        toggleCompany() {
            this.get('validateSync')();
        }
    }
});
```
This results in the following error:
```
Uncaught Error: Assertion Failed: Cannot call get with 'validations' on an undefined object
```

So how can I trigger a manual validation from within a component action?
Or is there a better way to do this?

offirgolan:
Resolves #104 

- Exposes `disabled` option for validators
- Add functionality to define `dependentKeys` in default options (should have already been there)

This will allow to do the following

```js

// Via default options
  username: {
    debounce: 500,
    disabled() { return !this.get('model._enableValidations'); },
    dependentKeys: ['_enableValidations'],
    validators: [
      validator('presence', true),
      validator('length', {
        max: 15
      })
    ]
  }
});

// Via single validator
  username: validator('presence', {
    presence: true,
    disabled() { return !this.get('model._enableValidations'); },
    dependentKeys: ['_enableValidations']
  })
});

  'username': attr('string'),
  _enableValidations: true
});
```

@stefanpenner @rwjblue open for comments.

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -42,7 +42,7 @@ export var validator = Validator;
  * }
  * from 'ember-cp-validations';
  *
- * var Validations = buildValidations({
+ * const Validations = buildValidations({
  *   username: validator('presence', true),
  *   password: [
  *     validator('presence', true),
@@ -91,7 +91,7 @@ export var validator = Validator;
  * }
  * from 'ember-cp-validations';
  *
- * var Validations = buildValidations({
+ * const Validations = buildValidations({
  *   bar: validator('presence', true)
  * });
  *
@@ -154,7 +154,7 @@ export var validator = Validator;
  * Instead of doing the following:
  *
  * ```javascript
- * var Validations = buildValidations({
+ * const Validations = buildValidations({
  *   username: [
  *     validator('presence', {
  *       presence: true,
@@ -174,7 +174,7 @@ export var validator = Validator;
  * We can declare default options:
  *
  * ```javascript
- * var Validations = buildValidations({
+ * const Validations = buildValidations({
  *   username: {
  *     description: 'Username'
  *     validators: [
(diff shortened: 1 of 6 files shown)


offirgolan:
ready for review :smiley_cat:

PR #112, opened 2016-02-05
Support validations via inheritance

### LINKED ISSUE #111: Validation on models with inheritance
Is there a 'best practice' for adding validations to models that are extending others?

For instance, if I have a Staff model that extends a User model, at the moment only the validations on the Staff model have any affect.

User model:
```javascript
import DS from 'ember-data';
import {validator, buildValidations} from 'ember-cp-validations';

    username: validator('presence', true),
    password: validator('presence', true)
});

    username: DS.attr('string'),
    password: DS.attr('string'),
    security_phrase: DS.attr('string'),
    last_login: DS.attr('date')
});
```

Staff model:
```javascript
import User from 'app/models/user';
import DS from 'ember-data';
import {validator, buildValidations} from 'ember-cp-validations';

    job_title: validator('presence', true)
});

export default User.extend(Validations, {
    job_title: DS.attr('string'),
    departments: DS.attr()
});
```

Here, only the job_title is validated. I've tried a few different ways, but they didn't really work, so at the moment Im just duplicating the validations in the child model.

offirgolan:
Resolves #111 

Allows validations to persist through inheritance

```js
const ParentValidations = buildValidations({...});
const Parent = Ember.Object.extend(ParentValidations);

const ChildValidations = buildValidations({...});
const Child = Parent.extend(ChildValidations);

var child = Child.create({...});

child.validateSync(); // should validate all props defined in parent and child
```


offirgolan:
ping @rwjblue :smile_cat:

addon/validations/factory.js
```diff
@@ -75,30 +76,27 @@ function setOwner(obj, model) {
...
-    validate,
-    validateSync
-  });
+  return Ember.Mixin.create({
+    validations: null,
 
-  return createMixin(GlobalValidations, AttrValidations);
+    init() {
+      this._super(...arguments);
+      this.set('validations', Validations.create({
```
stefanpenner: don't we usually want to do this on-demand via CP getter?

addon/validations/factory.js
```diff
@@ -133,94 +131,53 @@ function processDefaultOptions(validations = {}) {
...
-  props.message = computed('messages.[]', cycleBreaker(function() {
-    return get(this, 'messages.0');
-  }));
+function createValidationsObject(validations = {}) {
+  return Ember.Object.extend({
+    model: null,
+    attrs: null,
+    isValidations: true,
 
+    _validators: {},
```
stefanpenner: this state should likely be `instance` state not defined on the prototype

addon/validations/factory.js
```diff
@@ -133,94 +131,53 @@ function processDefaultOptions(validations = {}) {
...
+    _debouncedValidations: {},
```
stefanpenner: ^

addon/validations/factory.js
```diff
@@ -133,94 +131,53 @@ function processDefaultOptions(validations = {}) {
...
+    _validatableAttributes: [],
```
stefanpenner: ^

addon/validations/factory.js
```diff
@@ -133,94 +131,53 @@ function processDefaultOptions(validations = {}) {
...
+    _validationRules: {},
```
stefanpenner: ^

addon/validations/factory.js
```diff
@@ -133,94 +131,53 @@ function processDefaultOptions(validations = {}) {
...
+
+    init() {
+      this._super(...arguments);
+      let inheritedValidations = this.get('model.validations');
+      let validatableAttributes = Object.keys(validations);
+      let validationRules = validations;
+      let attrs = {};
+
+      if(inheritedValidations) {
+        validatableAttributes = emberArray(validatableAttributes.concat(getWithDefault(inheritedValidations, '_validatableAttributes', []))).uniq();
```
stefanpenner: is emberArray used just to get uniq? If so, we should avoid that

addon/validations/factory.js
```diff
@@ -133,94 +131,53 @@ function processDefaultOptions(validations = {}) {
...
+        validationRules = merge(merge({}, inheritedValidations.get('_validationRules')), validationRules);
```
stefanpenner: wouldn't super in a CP allow this to work on-demand like before?


```js
computed({
  get() { 
    let inherited = this._super(...arguments);

    return merge(current, inherited);
  }
})
```

addon/validations/factory.js
```diff
@@ -266,6 +223,67 @@ function createCPValidationFor(attribute, validations) {
...
+    isValidating: or(...validatableAttrs.map((attr) => `attrs.${attr}.isValidating`)).readOnly(),
+    isDirty: or(...validatableAttrs.map((attr) => `attrs.${attr}.isDirty`)).readOnly(),
+    isAsync: or(...validatableAttrs.map((attr) => `attrs.${attr}.isAsync`)).readOnly(),
+    isNotValidating: not('isValidating').readOnly(),
+    isInvalid: not('isValid').readOnly(),
+    isTruelyValid: and('isValid', 'isNotValidating').readOnly(),
+
+    messages: computed(...validatableAttrs.map((attr) => `attrs.${attr}.messages`), function() {
+      var messages = [];
+      validatableAttrs.forEach((attr) => {
```
stefanpenner: if we are already doing flatten bellow, this can just be a `map`

addon/validations/factory.js
```diff
@@ -266,6 +223,67 @@ function createCPValidationFor(attribute, validations) {
...
+
+    _promise: computed(...validatableAttrs.map((attr) => `attrs.${attr}._promise`), function() {
+      var promises = [];
+      validatableAttrs.forEach((attr) => {
+        var validation = get(this, `attrs.${attr}`);
+        if (get(validation, 'isAsync')) {
+          promises.push(get(validation, '_promise'));
+        }
+      });
+      return RSVP.all(flatten(promises));
```
stefanpenner: `RSVP.Promise.all`

addon/validations/factory.js
offirgolan: That's what it was before, but because of [this line](https://github.com/offirgolan/ember-cp-validations/pull/112/files#diff-795f2437a28d93a92495ebda9b421065R157), it goes into an infinite loop.

addon/validations/factory.js
offirgolan: These should be held on the proto as they act as a cache for all model validators

addon/validations/factory.js
stefanpenner: that is an empty line :trollface:

offirgolan: LOL it changed after I made a commit. See [here](https://github.com/offirgolan/ember-cp-validations/pull/112/files#diff-795f2437a28d93a92495ebda9b421065R160)

addon/validations/factory.js
```diff
@@ -133,94 +126,55 @@ function processDefaultOptions(validations = {}) {
...
 
-  props.message = computed('messages.[]', cycleBreaker(function() {
-    return get(this, 'messages.0');
-  }));
+    validate,
+    validateSync,
 
+    init() {
+      this._super(...arguments);
+      let inheritedValidations = this.get('model')._super();
```
offirgolan: @stefanpenner got it! This solved it. Im just not sure why this returns the `validations` object on that model and not the model itself....

offirgolan: Hmmm seems like this doesnt work for Ember >= 2.1

addon/validations/factory.js
```diff
@@ -133,94 +126,55 @@ function processDefaultOptions(validations = {}) {
...
 
-  props.message = computed('messages.[]', cycleBreaker(function() {
-    return get(this, 'messages.0');
-  }));
+    validate,
+    validateSync,
 
+    init() {
+      this._super(...arguments);
+      let inheritedValidations = this.get('model').constructor.superclass.prototype.get('validations');
```
offirgolan: This line is pretty sketchy. Before it was `this.get('model')._super()` but that doesnt work for Ember >= 2.1.

offirgolan:
@stefanpenner ready for final review

addon/validations/factory.js
```diff
@@ -133,94 +126,56 @@ function processDefaultOptions(validations = {}) {
...
-  props.messages = computed(...validatableAttrs.map((attr) => `attrs.${attr}.messages`), function() {
-    var messages = [];
-    validatableAttrs.forEach((attr) => {
-      var validation = get(this, `attrs.${attr}`);
-      if (validation) {
-        messages.push(get(validation, 'messages'));
-      }
-    });
+    // Caches
+    _validators: {},
```
stefanpenner: i think the caches want to be on the instance, not the prototype.

addon/validations/factory.js
```diff
@@ -133,94 +126,56 @@ function processDefaultOptions(validations = {}) {
...
+    _debouncedValidations: {},
 
-    return emberArray(flatten(messages)).compact();
-  });
+    // Private
+    _validationRules: null,
```
stefanpenner: same with these slots

addon/validations/factory.js
offirgolan: Caches were always on the proto. That's how we decided to write them. They are cached by model guid. See test [here](https://github.com/offirgolan/ember-cp-validations/blob/master/tests/integration/validations/factory-general-test.js#L411-L440)

addon/validations/factory.js
offirgolan: Arent these already on the instance? Does setting these to null still put them on the proto?

addon/validations/factory.js
offirgolan: You can see the [code for the cache](https://github.com/offirgolan/ember-cp-validations/blob/master/addon/validations/factory.js#L373-L382). If we remove it from the proto, there isnt really a need for this...

PR #123, opened 2016-02-17
Update docs to reflect correct arg order for createErrorMessage

kat3kasper:


offirgolan:
Whoops! Thanks for catching this :smile_cat:

PR #127, opened 2016-02-23
DS Error validator

### LINKED ISSUE #118: How to handleDS.errors?
Is there any code need be wriiten if server throws 422 errors.The one which is mentioned in the below link.

http://emberjs.com/api/data/classes/DS.Errors.html

offirgolan:
Creates a link between this library and Ember-Data's [DS.Errors](http://emberjs.com/api/data/classes/DS.Errors.html) to fetch the latest message for the given attribute.

Resolves #118 
 
```javascript
 validator('ds-error');
```


offirgolan:
@stefanpenner @rwjblue @blimmer open for review

app/validators/ds-error.js
```diff
@@ -0,0 +1,41 @@
+/**
+ * Copyright 2015, Yahoo! Inc.
```
blimmer: idk if you need to bump this up for legal stuffs

app/validators/ds-error.js
```diff
@@ -0,0 +1,41 @@
...
+ *  @class DS Error
+ *  @module Validators
+ *  @extends Base
+ */
+export default Base.extend({
+  validate(value, options, model, attribute) {
+    let errors = get(model, 'errors');
+
+    if(!isNone(errors) && errors instanceof DS.Errors && errors.has(attribute)) {
+      return get(errors.errorsFor(attribute), 'lastObject.message');
```
blimmer: I think it makes sense to get the `lastObject` in the errors array, but we should definitely document that behavior so people don't get confused? Do you think we need to expose a way to get all the errors?

app/validators/ds-error.js
offirgolan: Ooooh yeah, Ill do it in a separate commit. Thanks for the reminder :D

app/validators/ds-error.js
offirgolan: Look at the comment above. `...fetch the latest message for the given attribute.`

blimmer: Ah, I forgot we're using the autogen-ed docs. Looks good.

offirgolan: :+1:

PR #132, opened 2016-03-04
Add support for nested keys in validation rules

### LINKED ISSUE #131: nested keys to validate
I would like to be able to add validation at a form or component level.  It would be amazing if I could assign validations to nested attributes of objects e.g `user.firstName`.    I'm struggling to find a way to do this, any help would be much appreciated.

```js
import Ember from 'ember';

  'user.firstName': validator('presence', true),
  'user.lasName': validator('presence', true),
  'account.number': validator('presence', true),
});

export default Ember.Component.extend(Validations, {
  user: null,  /* object  { firstName: 'bob', lastName: 'billy' } */
  account: null, /* object { number: 123, balance: 456 } */

  isFormValid: Ember.computed.alias('validations.isValid'),
});
```

offirgolan:
Resolves #131 

```js
import Ember from 'ember';

  'user.firstName': validator('presence', true),
  'user.lasName': validator('presence', true),
  'account.number': validator('presence', true),
});

export default Ember.Component.extend(Validations, {
  user:  { firstName: 'bob', lastName: 'billy' } ,
  account: { number: 123, balance: 456 }

  isFormValid: Ember.computed.alias('validations.isValid'),
});
```

### DIFF
--- a/addon/validations/factory.js
+++ b/addon/validations/factory.js
@@ -6,6 +6,7 @@
 import Ember from 'ember';
 import getOwner from 'ember-getowner-polyfill';
 import flatten from '../utils/flatten';
+import assign from '../utils/assign';
 import ValidationResult from './result';
 import ValidationResultCollection from './result-collection';
 import BaseValidator from '../validators/base';
@@ -13,6 +14,7 @@ import cycleBreaker from '../utils/cycle-breaker';
 
 const {
   get,
+  set,
   run,
   RSVP,
   merge,
@@ -24,6 +26,7 @@ const {
   makeArray,
   canInvoke,
   getWithDefault,
+  defineProperty,
   A: emberArray
 } = Ember;
 
@@ -175,22 +178,48 @@ function createValidationsObject(validations = {}) {
     _validatableAttributes: null,
     _parentValidations: null,
 
+    validatableAttributes: computed.oneWay('_validatableAttributes'),
+
     validate,
     validateSync,
 
+    /**
+     * - Setup parent validation inheritance
+     * - Normalize nested keys (i.e. 'details.dob') into objects (i.e { details: { dob: validatior() }})
+     * - Merge normalized validations with parent
+     * - Create `attrs` object with CPs
+     * - Create global CPs (i.e. 'isValid', 'messages', etc...)
+     */
     init() {
       this._super(...arguments);
-      let validationRules = validations;
+      let validationRules = {};
       let parentValidations = this.get('_parentValidations');
-      let attrs = {};
+      let validatableAttributes = Object.keys(validations);
+      let model = this.get('model');
+      let attrs = Ember.Object.create();
 
+      // Setup validation inheritance
       if(parentValidations) {
-        validationRules = merge(merge({}, parentValidations.get('_validationRules')), validationRules);
+        validationRules = merge(validationRules, parentValidations.get('_validationRules'));
+        validatableAttributes = emberArray(parentValidations.get('_validatableAttributes').concat(validatableAttributes)).uniq();
       }
 
-      let validatableAttributes = Object.keys(validationRules);
+      // Normalize nested keys into actual objects and merge them with parent object
+      Object.keys(validations).reduce(( obj, key )=> {
+        assign(obj, key, validations[key]);
+        return obj;
+      }, validationRules);
+
+      // Create attrs object with CPs
       validatableAttributes.forEach((attribute) => {
-        attrs[attribute] = createCPValidationFor(attribute, validationRules[attribute]);
+        assign(attrs, attribute, createCPValidationFor(attribute, validationRules[attribute]));
+
+        // Add a reference to the model in the deepest object
+        let path = attribute.split('.');
+        let lastObject = get(attrs, path.slice(0, path.length  - 1).join('.'));
+        if(isNone(get(lastObject, '_model'))) {
+          set(lastObject, '_model', model);
+        }
       });
 
       this.setProperties({
@@ -198,9 +227,7 @@ function createValidationsObject(validations = {}) {
         _validationRules: validationRules,
         _validators: {},
         _debouncedValidations: {},
-        attrs: Ember.Object.extend(attrs).create({
-          _model: this.get('model')
-        })
+        attrs
       });
 
       createGlobalValidationProps(this);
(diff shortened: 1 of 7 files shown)


offirgolan:
@stefanpenner @rwjblue open for review

donovan-graham:
This looks great. Looking forward to it being merged.

addon/validations/factory.js
```diff
@@ -102,11 +104,15 @@ function setOwner(obj, model) {
 export default function buildValidations(validations = {}) {
   processDefaultOptions(validations);
 
-  const Validations = createValidationsObject(validations);
+  var Validations;
```
stefanpenner: `let`

addon/validations/factory.js
```diff
@@ -102,11 +104,15 @@ function setOwner(obj, model) {
...
 
   return Ember.Mixin.create({
     validations: computed(function() {
-      return Validations.create({ model: this, _parentValidations: this._super() });
+      if(!Validations) {
+        // We only want to create this once per class
+        Validations = createValidationsClass(this._super(), validations);
```
stefanpenner: I believe we may want to split this into 2 CP, `validationClass` and `validation`, as we may risk a memory leak by preserving the result of (or  member of) `_super()` which is an instance of the supers Validation class, rather then the class. The instance is allow to store instance specific state, and Im slightly worried that the `createValidationsClass` method may accidentally pass along some of that instance state.

Maybe something like:
```js
let Validations;
...
validationsClass: computed(function() {
  if (!Validations) {
     Validations = createVallidationClass(this._super(), validations)
  },
  
  return Validations
}),

validations: computed(function() {
   return this.get('validationClass').create({ model: this })
})
```

Thoughts?

addon/validations/factory.js
```diff
@@ -153,68 +159,105 @@ function processDefaultOptions(validations = {}) {
...
+  // Create the mixin that holds all the top level validation props (isValid, messages, etc)
+  const TopLevelProps = createTopLevelPropsMixin(validatableAttributes);
+
+  // Create the `attrs` class which will add the current model reference once instantiated
+  const Attrs = Ember.Object.extend(attrCPs, {
+    init() {
+      this._super(...arguments);
+      let model = this.get('_model');
+
+      validatableAttributes.forEach((attribute) => {
```
stefanpenner: rather then explicitly settnig `_model` on each of the validationAttributes, what if we only set it on the validation instance, and each validationAttribute is provided a reference to the `validation` on is instanstantiation, at which point the validationAttribute merely needs the following `cp : _model: Ember.comptued.readOnly('validation._model')`

stefanpenner: Maybe this isn't that importat...

addon/validations/factory.js
```diff
@@ -153,68 +159,105 @@ function processDefaultOptions(validations = {}) {
...
+        // Add a reference to the model in the deepest object
+        let path = attribute.split('.');
+        let lastObject = get(this, path.slice(0, path.length  - 1).join('.'));
+        if(isNone(get(lastObject, '_model'))) {
```
stefanpenner: is their a case where this will be invoked and the child already has a model?

addon/validations/factory.js
```diff
@@ -153,68 +159,105 @@ function processDefaultOptions(validations = {}) {
...
+    }
+  });
+
+  // Create `validations` class
+  return Ember.Object.extend(TopLevelProps, {
     model: null,
     attrs: null,
     isValidations: true,
 
+    validatableAttributes: computed.alias('_validatableAttributes').readOnly(),
```
stefanpenner: `computed.readOnly`

addon/validations/factory.js
```diff
@@ -153,68 +159,105 @@ function processDefaultOptions(validations = {}) {
...
+      let validatableAttrs = get(this, 'validatableAttributes');
       let debouncedValidations = get(this, `_debouncedValidations`);
 
       // Cancel all debounced timers
-      Object.keys(debouncedValidations).forEach(attr => {
-        let attrCache = debouncedValidations[attr];
-        // Itterate over each attribute and cancel all of its debounced validations
-        Object.keys(attrCache).forEach(v => run.cancel(attrCache[v]));
+      validatableAttrs.forEach(attr => {
+        let attrCache = get(debouncedValidations, attr);
```
stefanpenner: i think the attr here will always be a simple property access, as debounceValidations is just a pojo. We should likely avoid the extra get here,just do

```
let attrCache = debounceValidations[attr] ;
if (attrCache === undefined) { return; }
Object.keys(attrCache).forEach( ... )
```

addon/validations/factory.js
```diff
@@ -425,11 +468,11 @@ function getValidatorsFor(attribute, model) {
 function getDebouncedValidationsCacheFor(attribute, model) {
   var debouncedValidations = get(model, `validations._debouncedValidations`);
 
-  if (isNone(debouncedValidations[attribute])) {
-    debouncedValidations[attribute] = {};
+  if (isNone(get(debouncedValidations, attribute))) {
```
stefanpenner:
```js
if (debouncedValidations[attribute] === undefined) {

}
```

`get` and `isNone` here seem wasteful

addon/validations/factory.js
offirgolan: Makes sense to me. This will expose `model.validationsClass` as well. Maybe we should make it private. Ill give this a shot.

addon/validations/factory.js
offirgolan: Not sure I follow you here.

addon/validations/factory.js
offirgolan: if you have `foo.bar.baz` and `foo.bar.blah`, then it will prevent setting the model twice on `bar` but really, it doesnt need to be there.

addon/validations/factory.js
offirgolan: :+1:

addon/validations/factory.js
offirgolan: Get is needed since `attribute` can be a nested key

addon/validations/factory.js
offirgolan: `attr` actually can be a nested key

addon/validations/factory.js
stefanpenner: private yes

addon/validations/factory.js
stefanpenner: ah ok, i see.

addon/validations/factory.js
stefanpenner: ah ok

addon/validations/factory.js
stefanpenner: didn't realize that, seems good

PR #143, opened 2016-03-16
make format validator phone international

### LINKED ISSUE #141: format validator type `phone` does not recognize international numbers
Actually (almost) everything starting with a '+' and then just digits is a valid number.
This for example should definitly be a valid number: `+41123456789`.

luxferresum:
The format validator can now recognize international numbers from germany and switzerland and is extensible for more country codes.

Should fix #141

### DIFF
--- a/app/validators/format.js
+++ b/app/validators/format.js
@@ -55,8 +55,23 @@ const {
 export default Base.extend({
   regularExpressions: {
     email: /^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/,
-    phone: /^([\+]?1\s*[-\/\.]?\s*)?(\((\d{3})\)|(\d{3}))\s*[-\/\.]?\s*(\d{3})\s*[-\/\.]?\s*(\d{4})\s*(([xX]|[eE][xX][tT]?[\.]?|extension)\s*([#*\d]+))*$/,
+    // phone: /^([\+]?1\s*[-\/\.]?\s*)?(\((\d{3})\)|(\d{3}))\s*[-\/\.]?\s*(\d{3})\s*[-\/\.]?\s*(\d{4})\s*(([xX]|[eE][xX][tT]?[\.]?|extension)\s*([#*\d]+))*$/,
     url: /(?:([A-Za-z]+):)?(\/{0,3})[a-zA-Z0-9][a-zA-Z-0-9]*(\.[\w-]+)+([\w.,@?^=%&amp;:\/~+#-{}]*[\w@?^=%&amp;\/~+#-{}])??/,
+  },
+
+  countyPhoneRegularExpressions: {
+    '+1': { // NANP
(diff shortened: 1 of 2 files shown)


offirgolan:
Hi @luxferresum, thanks for taking the time in putting this PR together. As of right now I dont think this is necessary just because it seems that your use-case is pretty specific to your region. The format validator regex for phone, email, and url and taken from credible sources and are will never meet everyone's requirements. Unless we can extend the current regex to support a general format for international numbers, I suggest making a custom validator to support your case by extending the format validator.

luxferresum:
I strongly disagree @offirgolan!
The current regex might be from a credible source but is definitely the one that is very specific to a region! It is only correct if you are both in the NANP area and only want to validate NANP numbers.

My PR changes this by allowing both, individual regex for every region and specifying in what region you are to allow local numbers.

Most important that it adds extensibility to support every country format.

Its important to understand that in international standards phone numbers can only contain digits and maybe a '+' and the start. This is the *only* requirement, and would result in a very simple regex. However this will not work for some regions where its common to write additional stuff next for a phone number, including NANP where its common to write an extension next to the phone number without standardized separator. This however makes it extremely hard to automatic process and/or validate phone numbers.

So probably the best regex that will work in most regions for most cases is `^\+?[0-9]( ?[0-9])*$`!

However this allows not to filter region specific limitations, like the 10 digit requirement in the NANP area.

Also it must be noticed that not only the local number format differs from region to region, also the international dialout is different in some regions.

While most countries of the world use `00` as international trunk access code, in NANP is it `011`, and in a few countries it is not standardized at all.
However in the GSM standard the `+` is an always working prefix for international dialout, so probably the only exact way to write down phone numbers is with a leading `+`. If you don't do this you can never be sure if you have a local or an international number.

The rest of the phone number usually has the only limitation that it only contain digits and in most countries it must start with a trunk prefix which is `0` in most countries. If the international prefix is specified the trunk prefix is usually omitted, so for example `0151 11111111` becomes `+49 151 11111111`.

Again there are exceptions to this, and again its NANP as well as for example Italy.
In NANP the is much thing as a trunk prefix, and in italy the leading `0` won't be omitted for international dialin.


As you see the probably only correct way to implement international phone number validation is to specify a international and a local regex for every region and enforce the leading `+` for the international regex. So the specialities of every country can be validated with a different regex.

blimmer:
This is an excellent, well thought out PR.

That said, I think this logic is too advanced for a base validator that ships to every person using cp-validations. As @offirgolan mentioned, the phone number validator comes from a reputable source and seems to meet the minimum viable for this library. The base validators are meant to work in the most general cases, with the flexibility to create custom validators as needed.

I think others could definitely benefit from this functionality outside of the core project. What do you think about creating an addon like `ember-cp-validations-intl` (or something similar) that provides this custom validator for folks who need this more advanced logic? Additionally, others could easily open PRs to that addon to add support for more countries, as you mentioned above.

luxferresum:
My version will not break any code but the current version will *not* work in most cases. It will not work for anyone outside of the NANP area.

blimmer:
I definitely understand what you're saying but @offirgolan and I would like to keep this validation very slim. Perhaps our inclusion of a phone validator in the first place was not the best idea because of the inherent complexity associated with phone numbers worldwide.

I really think the addon is the best way to go with this, and it would be quite simple to get working.

offirgolan:
@luxferresum I completely understand the need for this addition but I must reiterate that it is far too complex for the general purpose of this library. In the current state of this addon, we expose numerous ways to provide your own modifications to a validator (i.e. extending, custom options, custom validators) to fit your own personal need. As said by @blimmer, this would make a great 3rd party addon which I would gladly put in the docs as I did for ember-intl-cp-validations and ember-i18n-cp-validations.

PR #146, opened 2016-03-18
Global options support

offirgolan:
If you have  specific options you want to propagate throught all your validation rules, you can do so by passing in a global options object.
This is ideal for when you have a dependent key that each validator requires such as the current locale from your i18n implementation, or
you want easily toggle your validations on/off.

```javascript
const Validations = buildValidations(validationRules, globalOptions);
```

```javascript
import Ember from 'ember';
  firstName: {
    description: 'First Name'
    validators: [
      validator('presence', {
        presence: true,
        dependentKeys: ['foo', 'bar']
      })
     ]
   },
  lastName: validator('presence', true)
}, {
  description: 'This field'
  dependentKeys: ['i18n.locale', 'disableValidations'],
  disabled() {
    return this.get('model.disableValidations');
  }
});
```

Just like in the default options, locale validator options will always take precedence over default options and default options will always take precedence
over global options. This allows you to declare global rules while having the ability to override them in lower levels.
This rule does not apply to `dependentKeys`, instead they all are merged. In the example above, __firstName__'s dependentKeys will be
`['i18n.locale', 'disableValidations', 'foo', 'bar']`

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -2,3 +2,4 @@ languages:
   JavaScript: true
 exclude_paths:
 - "tests/*"
+- "config/*"
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ env:
 matrix:
   fast_finish: true
   allow_failures:
+    - env: EMBER_TRY_SCENARIO=ember-beta
     - env: EMBER_TRY_SCENARIO=ember-canary
 
 before_install:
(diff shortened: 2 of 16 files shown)


offirgolan:
Im not too sure if this requires a major version upgrade or not since there was a minor api change.

From:

```js
buildOptions(options = {}, defaultOptions = {}})
```

To:

```js
buildOptions(options = {}, defaultOptions = {}, globalOptions = {})
```

Which means that anyone that has a custom validator which uses `buildOptions` will have to make this change.

PR #152, opened 2016-03-28
export validators from addon namespace

luxferresum:
move the validators to the addon namespace an just reflect them in the app namespace for better extensibility.


addon/validators/belongs-to.js
```diff
@@ -0,0 +1,93 @@
+/**
+ * Copyright 2016, Yahoo! Inc.
+ * Copyrights licensed under the New BSD License. See the accompanying LICENSE file for terms.
+ */
+
+import Ember from 'ember';
+import Base from 'ember-cp-validations/validators/base';
```
offirgolan: Can you change all of these to:

```js
import Base from './validators/base';
```

luxferresum:
Do you mean `./base`?

    import Base from './base';

I can do this but I don't think full path is bad.

offirgolan:
Yes

On Monday, March 28, 2016, Lukas Kohler <notifications@github.com> wrote:

> Do you mean ./base?
>
> import Base from './base';
>
> I can do this but I don't think full path is bad.
>
> —
> You are receiving this because you commented.
> <https://github.com/offirgolan/ember-cp-validations/pull/152#issuecomment-202641639>
>


-- 
Thanks,
Offir Golan

luxferresum:
Well, I've done that with cff938b

offirgolan:
:+1: Thank you :smile_cat:

PR #161, opened 2016-04-13
Alias validator

### LINKED ISSUE #160: `alias` validator
Recently, I found myself needing an `alias` validator. It copies first error message from an `aliased` field to another it is bound to (similar to how `ds-error` copies error from `model.errors.${attr}`). 

Internally it's very much like  `dependent` validator because it depends on  `validations.attrs.${alias}.isTruelyValid`, but unlike `dependent` it copies error message from
`validations.attrs.${alias}.messages.firstObject` 

Intended usage:
```
{
    date: [
        validator('ds-error')
    ],
    dateFormatted: [
        validator('alias', {
            alias: 'date',
        })
    ]
}
```

The example use case I had was: 
UI form is working on computed property named `dateFormatted`, which depends on `date`. 
Raw value of `date` is never displayed nor edited directly by the user. 
When sending payload to server, `date` is sent and all errors returned by server are assigned to `date` model attribute and those should be displayed under `dateFormatted`.

Not sure, how common is my use case, but I would like to share how I wanted to solve this, how I actually solved that (see next comment), and maybe it would turn out that such validator could be built into cp-validations.

offirgolan:
Creates an alias between a single attribute's validations to another. 

 ```javascript
 // Examples
 validator('alias', 'attribute')
 validator('alias', {
   alias: 'attribute'
 })
 ```

@blimmer Let me know what you think. I went kinda crazy and cleaned up some stuff.

The important files are: 

- `addon/validations/factory.js`
- `addon/validators/alias.js`
- `tests/integration/validations/model-relationships-test.js`

Resolves #160


tests/integration/validations/model-relationships-test.js
```diff
@@ -230,6 +236,65 @@ test("belongs-to relationship returns undefined", function(assert) {
...
+  })));
+
+  user.get('validations').validateSync();
+
+  assert.equal(user.get('validations.isValid'), false);
+  assert.equal(user.get('validations.isValidating'), false);
+  assert.equal(user.get('validations.attrs.firstName.isValid'), false);
+  assert.equal(user.get('validations.attrs.lastName.isValid'), false);
+  assert.equal(user.get('validations.attrs.fullName.isValid'), false);
+  assert.deepEqual(user.get('validations.attrs.fullName.messages').sort(), ['firstName should be present', 'lastName should be present'].sort());
```
blimmer: nice - i think this is the right behavior vs. `firstObject`.

blimmer:
This LGTM cc @gustaff-weldon

offirgolan:
I feel like there are will be a use-case where you might not want to
include every error message in the aliases validation. What do you think of
adding a "firstMessageOnly" option or something like that?

On Sunday, April 17, 2016, Ben Limmer <notifications@github.com> wrote:

> This LGTM cc @gustaff-weldon <https://github.com/gustaff-weldon>
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/offirgolan/ember-cp-validations/pull/161#issuecomment-210937463>
>


-- 
Thanks,
Offir Golan

blimmer:
I think the option is reasonable, or just documentation around `firstObject`. It's a pretty common ember pattern to use to get the first object from an array.

offirgolan:
@blimmer not sure I follow what you mean.

blimmer:
Sorry, just reread my comment and it wasn't very clear. I was thinking that a `firstMessageOnly` will really just get the messages and then call `.firstObject` on the array. I think most ember folks know about it, but I think the option is a good idea for validations that you always want only one message. I could see people using that as a global validation option, too, so that plan sounds good to me!

PR #164, opened 2016-04-14
Replaces Ember logo with registered one

locks:


offirgolan:
:+1: thank you :smile_cat:

PR #167, opened 2016-04-14
Add null safe checks for relational validators after promise resolves

### LINKED ISSUE #162: Nested belongs-to validators gives error
Hey,

I've got nested `belongs-to` validators and the relationship does not have to exist. Eg.

```javascript
//model/foo.js
var Validations = buildValidations({
    bar: validator('belongs-to')
});

    name: DS.attr(),
    bar: DS.belongsTo('bar')  
});

//model/bar.js
var Validations = buildValidations({
    baz: validator('belongs-to')
});

    name: DS.attr(),
    baz: DS.belongsTo('baz')
});

//model/baz.js
var Validations = buildValidations({
    name: validator('presence', true)
});

    name: DS.attr()  
});
```

Creating a `foo` model and calling validate without adding `bar` and `baz` now triggers and error.

```javascript

let model = this.store.createRecord('foo');
model.validate({}, true).then(({model, validations}) => {
    console.log(model, validations); 
});
``` 

gives

```
ember.debug.js:32127 Error: Assertion Failed: Cannot call get with 'validations' on an undefined object.
    at new Error (native)
    at Error.EmberError (http://localhost:4202/assets/vendor.js:26688:21)
    at assert (http://localhost:4202/assets/vendor.js:16447:13)
    at Object.assert (http://localhost:4202/assets/vendor.js:26452:34)
    at get (http://localhost:4202/assets/vendor.js:30911:22)
    at http://localhost:4202/assets/vendor.js:69404:20
    at tryCatch (http://localhost:4202/assets/vendor.js:63616:14)
    at invokeCallback (http://localhost:4202/assets/vendor.js:63631:15)
    at publish (http://localhost:4202/assets/vendor.js:63599:9)
    at http://localhost:4202/assets/vendor.js:42155:7
``` 

I've traced it to the `belongs-to` validator and temporarily added a check for the model variable:

```javascript

export default Base.extend({
  validate(value) {
    if (value) {
      if (canInvoke(value, 'then')) {
        return value.then(model => {
            return model && get(model, 'validations');
        });
      }
      return get(value, 'validations');
    }

    return true;
  }
});

```
Do you have any input on how to solve this? Is there any other way to define the nested `belongs-to`?

Cheers,
Carl

offirgolan:
Resolves #162

### DIFF
--- a/addon/validators/belongs-to.js
+++ b/addon/validators/belongs-to.js
@@ -83,7 +83,7 @@ export default Base.extend({
   validate(value) {
     if (value) {
       if (canInvoke(value, 'then')) {
-        return value.then(model => get(model, 'validations'));
+        return value.then(model => model ? get(model, 'validations') : true);
       }
       return get(value, 'validations');
     }
--- a/addon/validators/has-many.js
+++ b/addon/validators/has-many.js
@@ -59,7 +59,7 @@ export default Base.extend({
   validate(value) {
     if (value) {
       if (canInvoke(value, 'then')) {
-        return value.then(models => models.map(m => m.get('validations')));
+        return value.then(models => models ? models.map(m => m.get('validations')) : true);
       }
       return value.map(m => m.get('validations'));
     }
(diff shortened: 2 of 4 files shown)


blimmer:
nice work and test coverage 😸

offirgolan:
😺

PR #170, opened 2016-04-16
Only call super on validations class if it exists via shouldCallSuper

### LINKED ISSUE #149: Asking for `validations` triggers save in Mixin
```
export default Ember.Component.extend(Validations, ComponentCRUD, {
  actions: {
    save() {
      const validations = this.get('validations'); //remove this line and won't hit debugger
   }
```

```
var ComponentCRUD = Ember.Mixin.create({
  actions: {
    save(model){
      debugger;
      this.attrs.save(model);
    },
....
```

Is this the desired behavior?

offirgolan:
All credit goes to @rwjblue for thinking this up. (Savage)

Resolves #149

### DIFF
--- a/addon/validations/factory.js
+++ b/addon/validations/factory.js
@@ -11,6 +11,7 @@ import ValidationResult from './result';
 import ValidationResultCollection from './result-collection';
 import BaseValidator from '../validators/base';
 import cycleBreaker from '../utils/cycle-breaker';
+import shouldCallSuper from '../utils/should-call-super';
 
 const {
   get,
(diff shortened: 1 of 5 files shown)


addon/validations/factory.js
```diff
@@ -98,14 +99,27 @@ function buildValidations(validations = {}, globalOptions = {}) {
   let Validations;
 
   return Ember.Mixin.create({
-    _validationsClass: computed(function () {
+    init() {
+      this._super(...arguments);
+
+      // Count number of mixins to bypass super check if there is more than 1
+      this.__validationsMixinCount__ = this.__validationsMixinCount__|| 0;
+      this.__validationsMixinCount__++;
```
rwjblue: Small tweak here, we should store the resulting integer in a closure scoped variable:

```js
let Validations, validationMixinIndex;
// ...snip...

validationMixinIndex = this.__validationsMixinCount__++;
```

Then below, the guard for calling super in `__validationsClass__` would look like:

```js
if(shouldCallSuper(this, '__validationsClass__') || validationMixinIndex > 1) {
  inheritedClass = this._super();
}
```

This ensures that the first mixin with a `__validationsClass__` only calls `_super` if the superclass passes the `shouldCallSuper` check, and the second (and any more) mixins always call `_super`.

offirgolan: :+1: Added the fix

PR #171, opened 2016-04-18
Value option

### LINKED ISSUE #166: Feature: Allowing attribute to be specified in options
We ran into a situation where we would like the validator to be tied to a specific property on the model, but we need the value passed into the validator to be based on a different related property. I didn't see a way in the code that would allow today, so I wanted to throw the idea out there.

A example could be something like (they must make 3 selections)

```
  selections: {
    description: "Order Selections",
    validators: [
      validator('ds-error'),
      validator('has-many'),
      validator('number', {
        lte: 3,
        attribute: 'selections.length'
      })
    ]
  }
})
```

This would allow us to group validation errors to display in the UI to be based on a single property, but allow the value passed into some of the validators to be based on a different attribute than the main attribute.

Is there way to do this today, if not, something you would be open to?

offirgolan:
Used to retrieve the value to validate. This will overwrite the validator's default `value` method.
By default this returns `model[attribute]`.

```javascript
// Examples
validator('date', {
   value() {
 	// Format the orignal value before passing it into the validator
 	const rawValue = this.get(this.get('model'), this.get('attribute'));
 	return moment().utc(rawValue).format('DD/MM/YYY');
   }
})
validator('number', {
   value() {
 	// Validate a value that is not the current attribute
 	return this.get('model').get('someOtherAttr');
   }
})
```

@blimmer @stefanpenner Let me know what you think of this. Implementation is open for debate.

Resolves #166


blimmer:
I really like the direction of this, especially with the date example. I have run into this exact situation quite a few times. My only question is about dependent keys. Is it the consumer's responsibility to pass additional dependent keys if they're relied upon in `value()`?

offirgolan:
Unfortunately yes. There is no way for us to determine what dependents are
needed.

On Friday, April 22, 2016, Ben Limmer <notifications@github.com> wrote:

> I really like the direction of this, especially with the date example. I
> have run into this exact situation quite a few times. My only question is
> about dependent keys. Is it the consumer's responsibility to pass
> additional dependent keys if they're relied upon in value()?
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/offirgolan/ember-cp-validations/pull/171#issuecomment-212889562>
>


-- 
Thanks,
Offir Golan

PR #173, opened 2016-04-20
Introduce `ignoreWhitespace` in presence validator

krasnoukhov:
This allows using a validator for a case when value should not consist only of space(s).

### DIFF
--- a/addon/validators/presence.js
+++ b/addon/validators/presence.js
@@ -8,7 +8,8 @@ import Base from './base';
 
 const {
   get,
-  isEmpty
+  isEmpty,
+  isPresent
 } = Ember;
 
 /**
@@ -21,6 +22,13 @@ const {
  *  validator('presence', {
  *    presence: true,
  *    message: 'should not be empty'
+ *  })
+ *
+ *  If you want whitespace to be ignored, use `ignoreWhitespace: true`:
+ *
+ *  validator('presence', {
+ *    ignoreWhitespace: true,
+ *    message: 'should not be empty or consist only of spaces'
  *  })
  *  ```
  *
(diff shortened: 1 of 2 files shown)


aaronbhansen:
I feel like this should be a configuration option for the presence validator. Something like 

```validator('presence', { ignoreWhitespace: true }) ```

or 

```validator('presence', { ignoreBlank: true }) ```

Since this would change existing functionality for people already using it today.

krasnoukhov:
Makes sense, thanks. I've updated the commit. CI is failing though, but it does not look like my change broke it, so not sure what can I do

krasnoukhov:
CI is green now!

offirgolan:
@krasnoukhov thank you for this feature! I think I prefer "ignoreWhitespace" more just because it's a little more intuitive, but I'm not sure if that's more misleading than ignoreBlank. May I ask why you chose ignoreBlank? Besides that, if you can just add the option details in the docs (see addon/validators/format), I believe I can just go ahead and merge this baby 😸

krasnoukhov:
@offirgolan No problem! Well `ignoreBlank` is just shorter, but I've updated a commit with a new name and polished doc.

offirgolan:
Oh okay I was just curious. After some thought, I believe ignoreWhitespace
sort of implies that it will ignore all whitspace characters in a string.
Maybe ignoreBlank was the better choice but regardless, I can always make
that change after I merge 😊. I'm currently on vacation so I'll merge this
along with some open PRs early next week and issue a release.

On Thursday, April 21, 2016, Dmitry Krasnoukhov <notifications@github.com>
wrote:

> @offirgolan <https://github.com/offirgolan> No problem! Well ignoreBlank
> is just shorter, but I've updated a commit with a new name and polished doc.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/offirgolan/ember-cp-validations/pull/173#issuecomment-212604585>
>


-- 
Thanks,
Offir Golan

krasnoukhov:
Sure, I don't mind at all, maybe you'll think of even better name for that! Thanks and have a good rest of the week.

blimmer:
LGTM

PR #174, opened 2016-04-22
Creating inline validator

aaronbhansen:
This allows you to create a validator inline that validates with a custom function. 

This will help when you need to create a one off validation without having to create a custom validator.

### DIFF
--- a/tests/unit/validators/date-test.js
+++ b/tests/unit/validators/date-test.js
@@ -217,7 +217,8 @@ test('after now or on', function(assert) {
   assert.expect(3);
   var now = moment().format('MMM Do, YYYY');
   options = {
-    onOrAfter: 'now'
+    onOrAfter: 'now',
+    percision: 'second'
   };
 
   message = validator.validate('1/1/2014', assign({}, options));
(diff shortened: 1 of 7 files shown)


tests/unit/validators/date-test.js
```diff
@@ -217,7 +217,8 @@ test('after now or on', function(assert) {
   assert.expect(3);
   var now = moment().format('MMM Do, YYYY');
   options = {
-    onOrAfter: 'now'
+    onOrAfter: 'now',
+    percision: 'second'
```
aaronbhansen: I noticed the milliseconds would randomly fail, when both values were now. This fixed it.

offirgolan:
This is already supported via validator(function(...))

aaronbhansen:
Oh nice, not sure how I missed that. 

Let me know if you want a pull request for the date test fix, or if you want to just add it in.

offirgolan:
If you can create a quick pr for it, I can merge it. Why was it failing for milliseconds and not seconds?

offirgolan:
And no worries 😸 Thanks for taking the time to put this together!

PR #177, opened 2016-04-26
Expose dependent keys api to validators

offirgolan:
This is a feature that has been a long time coming and requested by many power users. This allows a user to declare custom dependent keys on the validator level for custom validators. This will remove a lot of duplicate code and will also reduce the validation rules pojo. 

Since we only create the `ValidationsClass` once and only do it on first get of the `validations` CP, we can just create the CPs on init which gives access to the container. I made the API a static function so we dont have to actually instantiate the validator to access the dependents. I was hoping to get feedback before I work on documentation.

Another thing I can add is

```bash
ember g validator --with-dependents
```

which will create a validator file with the static function override via `reopenClass`.

Ping @stefanpenner @rwjblue @blimmer 😸

### DIFF
--- a/addon/validations/factory.js
+++ b/addon/validations/factory.js
@@ -180,13 +180,13 @@ function normalizeOptions(validations = {}, globalOptions = {}) {
  *   - Setup parent validation inheritance
  *   - Normalize nested keys (i.e. 'details.dob') into objects (i.e { details: { dob: validator() }})
  *   - Merge normalized validations with parent
- *   - Create `attrs` object with CPs
  *   - Create global CPs (i.e. 'isValid', 'messages', etc...)
  *
  * @method createValidationsClass
  * @private
  * @param  {Object} inheritedValidationsClass
  * @param  {Object} validations
+ * @param  {Object} owner
  * @return {Ember.Object}
  */
 function createValidationsClass(inheritedValidationsClass, validations, owner) {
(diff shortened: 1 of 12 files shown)


addon/validations/factory.js
```diff
@@ -207,21 +207,21 @@ function createValidationsClass(inheritedValidationsClass, validations = {}) {
...
-  const Attrs = Ember.Object.extend(attrCPs, {
+  const Attrs = Ember.Object.extend({
     init() {
       this._super(...arguments);
       const model = this.get('_model');
 
+      // Create the CPs
+      validatableAttributes.forEach(attribute => {
+        const cp = createCPValidationFor(attribute, get(validationRules, attribute), owner);
+        assign(this, attribute, cp, true);
```
stefanpenner: This is a feature that only exists due to a bug in ember, lets not rely on it. Basically there is no way currently to performently add CP's to instances, only classes. I have toyed around with the idea of adding support for adding instance specific dependentKeys to existing CP's, but it hasn't be implemented yet. Would that help?

offirgolan: Oh man, I've been gone for a week and I already forgot we spoke about this. I feel like something like this will be very handy so we should def. explore a more optimal path but this is already being done in the current version of this library via the `assign` util method. When you have nested keys, it creates a new Ember.Object for each nested key, and at the end, uses `defineProperty` on the instance. See [here](https://github.com/offirgolan/ember-cp-validations/blob/master/addon/utils/assign.js#L37).

If you end up with a viable solution, im assuming it would only work for the latest ember version and above. Would we be able to back port such a feature?

blueprints/validator/files/app/validators/__name__.js
```diff
@@ -1,7 +1,22 @@
...
+   * @param {String}  attribute   The attribute being evaluated
+   * @param {Unknown} options     Options passed into your validator
+   * @return {Array}
+   */
+  getDependentsFor(/* attribute, options */) {
+    return [];
+  }
+});
+
+export default <%= classifiedModuleName %>;
```
offirgolan: @blimmer Thoughts about this new default validator template? Does it feel too complicated? I decided to expose this so anyone creating a new validator will know its there.

addon/validations/factory.js
offirgolan: Noting that after discussion with @stefanpenner, using `defineProperty` is okay since it is only being done once _**per class**_ and not _per instance_

blueprints/validator/files/app/validators/__name__.js
blimmer: I don't think this is too complex, personally.

Issue #113, opened 2016-02-06
Feature Request: hasValidated as validation state


jasonmit:
A common pattern is surfacing validation errors after the user touched a field.

It would be great if this state was stored on the validation object and flipped only when `validate` or `validateSync` is invoked and in the case of the async validation - once it's resolved.

Thoughts?  I can work on this if there is buy-in.

offirgolan:
Im not exactly sold on the idea that this should be part of this library. I feel like we provide enough state properties such as `isValidating` and `isValid` (or `isTruelyValid`), that this type of state should be handled solely by the user. 

```js
model.validate().then(() => {
   // do some complex stuff here
  model.set('_hasValidated', true); 
  model.save();
});
```

I think that the term `hasValidated` is sort of misleading, in the sense that it implies that calling `validate` or `validateSync` actually trigger validations where actually, validations happen automatically via the CPs.

jasonmit:
> validations happen automatically via the CPs.

I see everyone having to manage this state https://github.com/offirgolan/ember-cp-validations/blob/bf1e4642d0358fe41f1bee550004831bc4b89a45/tests/dummy/app/controllers/index.js#L33

which I think can be internalized since you know when validate/validateSync was invoked.  Dirtying the any of the attributes would flip this state back.  Closing and will implement it on my end.

offirgolan:
Hmmmmm I see your point. So this leads to two questions:

1. When is this flag turned on
2. When/how is this flag turned off

jasonmit:
> When is this flag turned on

After `validate` or `validateSync` are invoked.

> When/how is this flag turned off

When an attribute that has a validator is touched.

What I'm saying is just a proposal, so feel free to keep bouncing ideas off me or whoever until you're comfortable with the idea.

offirgolan:
Maybe the best way to go about this is (if you have some spare time) to create this feature and open a PR and we can discuss from there :smile_cat:

acburdine:
I'll see what I can do sometime this week on this.

v-shan:
Hi all! Any updates here? The `didValidate` property would be very useful for us. Right now to get that functionality, the options are:

- use `targetObject` to get the `didValidate` property (seems brittle) or
- pass the `didValidate` property in (seems unnecessary)

In both cases we have to pass `didValidate` or `targetObject` down a component chain if there are nested components, which is also :(. 

For reference, our use case:
[...]

PR #182, opened 2016-05-02
[WIP] hasValidated property 

acburdine:
refs #113 

@offirgolan this is a result of the little bit of time I was able to spend on this, it doesn't work very well at the moment because I'm still getting used to the codebase but will see what all I can do tonight/later this week.

### DIFF
--- a/addon/validations/factory.js
+++ b/addon/validations/factory.js
@@ -641,6 +641,8 @@ function validate(options = {}, async = true) {
     content: validationResults
   });
 
+  validationResultsCollection.setValidated();
+
   const resultObject = {
     model,
     validations: validationResultsCollection
--- a/addon/validations/result-collection.js
+++ b/addon/validations/result-collection.js
@@ -16,6 +16,7 @@ const {
   isEmpty,
   isArray,
   isNone,
+  canInvoke,
   A: emberArray
 } = Ember;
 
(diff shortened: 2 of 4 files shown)


acburdine:
@offirgolan I've managed to get this working for everything but the hasMany validator; however, the current handling of belongsTo validators isn't ideal, and I may need some guidance as to a better way to accomplish this 😕

offirgolan:
@acburdine super sorry for such a delayed response! Thank you so much for taking the time to put this together. 

I feel as though this is overly complicated, but then again, my use-cases in the past have been very simple. Why is there a need to reset `hasValidated` when a value changes? Maybe `hasValidated` isnt the correct name for this. 

//cc @jasonmit

acburdine:
@offirgolan The reason hasValidated is reset when the value changes is because of [this comment](https://github.com/offirgolan/ember-cp-validations/issues/113#issuecomment-181717343).

Personally, my use case is for only showing the error once the input has focused out, which means whenever the `validate` or `validateSync` method is called. I'm not sure what a better name for it would be though 😕

offirgolan:
Haha okay makes sense 😸 . Let me take some time to think about this and get back to you. I feel like this solution can either be simplified or can be split into a `didValidate` and a `hasChanged` sorta thing.

offirgolan:
Im gonna go ahead and close this PR as it is pretty old. @acburdine thanks again for taking the time to put this PR together but as of right now, I think the simplest solution is to track this state on the app-side.

PR #189, opened 2016-05-20
Fix syntax error in phone validator example

mike-north:


offirgolan:
Thanks 😛

mike-north:
My meaningful contribution to this library

PR #207, opened 2016-06-03
Bug fixes

offirgolan:
- null length should error
- add parentAttribute in errors
- add .[] dependent key in has-many
- extend dependents from base validator

### DIFF
--- a/addon/validations/error.js
+++ b/addon/validations/error.js
@@ -15,6 +15,13 @@ export default Ember.Object.extend({
   attribute: null,
 
   /**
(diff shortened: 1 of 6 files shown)


aalimovs:
`Null or undefined value in length validator should be invalid`

I assume ember-cp-validations follows semver. How is this not a breaking change?

offirgolan:
@aalimovs that was my mistake. Ill add upgrade notes to the release. Please use `allowBlank` to allow null or undefined values.

PR #215, opened 2016-06-15
Add import for getOwner to README example

jrjohnson:
Needed to specify that this function comes out of the Ember object in the README example docs.


offirgolan:
Ah thanks for catching that! 😸

PR #218, opened 2016-06-21
Fix for data validator with format

patricklx:
The date validator is not working with format specified.
It always says invalid value.

It is first validated without the format, returns false, since Date() cannot parse it and then will return an error message and never tries with the format.
This fix skips the Date() check if format is specified
and fixes the data validator test to use a format that is not parsed by
Date()

### DIFF
--- a/addon/validators/date.js
+++ b/addon/validators/date.js
@@ -63,21 +63,21 @@ export default Base.extend({
     const format = options.format;
     const precision = options.precision;
     let { before, onOrBefore, after, onOrAfter } = options;
+    let date;
 
     if (options.allowBlank && isEmpty(value)) {
       return true;
-    }
-
-    let date = this._parseDate(value);
-
-    if (!date.isValid()) {
-      return this.createErrorMessage('date', value, options);
-    }
+    }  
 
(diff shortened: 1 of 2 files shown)


addon/validators/date.js
```diff
@@ -68,10 +68,12 @@ export default Base.extend({
       return true;
     }
 
-    let date = this._parseDate(value);
-
-    if (!date.isValid()) {
-      return this.createErrorMessage('date', value, options);
+    let date;
```
offirgolan: Can you move this to the top of the function after `let { before ... }`

addon/validators/date.js
```diff
@@ -68,10 +68,12 @@ export default Base.extend({
...
+    if (!format) {
```
offirgolan: Move this to the `else` block of `if(format)`

offirgolan:
@patricklx thank you catching this! I made a couple comments, once they get resolved I will merge.

patricklx:
done ;)

offirgolan:
Perfect! Thanks again 😺

PR #225, opened 2016-06-23
Use Require for Checking Ember Data

### LINKED ISSUE #224: Ember-Data is required to use the DS Error validator - Too Strict for ED 2.6+
In recent versions of Ember Data, validating that Ember Data exists by `self.DS` is no longer effective - because ED imports work like this now:
```
import attr from 'ember-data/attr';
```

This error is thrown:
https://github.com/offirgolan/ember-cp-validations/blob/d682193c1bb978ad56b300d38c1297e42aa96ba7/addon/validators/ds-error.js#L10

I'd recommend something like this:
```
try {
  requireModule('ember-data');
} catch {
  throw new Error('Ember-Data is required to use the DS Error validator.');
}
```

That cool?  if so I'll shoot over a PR

hhff:


offirgolan:
@hhff JSHint tests are failing. `modules/ember-cp-validations/validators/ds-error.js: line 36, col 46, 'DS' is not defined.` 

Im not really sure what the best way to proceed would be. Maybe @stefanpenner or @rwjblue can give some direction here.

addon/validators/ds-error.js
```diff
@@ -5,11 +5,11 @@
...
 import Ember from 'ember';
 import Base from 'ember-cp-validations/validators/base';
-const DS = self.DS;
+import require from 'require';
 
-if (typeof self.DS === 'undefined') {
-  throw new Error('Ember-Data is required to use the DS Error validator.');
-}
+const { assert } = Ember;
+const hasEmberData = require.has('ember-data');
```
offirgolan: I would prefer to put this logic in [addon/utils/utils.js](https://github.com/offirgolan/ember-cp-validations/blob/master/addon/utils/utils.js#L9)

hhff: 👍 will do

dustinfarris:
I think either make ember-data a hard dependency or move ds-error to a separate addon.

offirgolan:
@dustinfarris I think such an extreme is required. Ill figure out a working solution to this.

dustinfarris:
ok.  also, maybe related, i forced the DS import in my local fork, and now it looks like `validate` is not getting fired when the model.save promise rejects

offirgolan:
Closing in favor of #239

hhff:
👍 thanks all!  ran out of time on this

offirgolan:
@hhff Np! Thanks for bringing this to our attention.

PR #226, opened 2016-06-23
[Feature] Warning Validators

offirgolan:
Use Case:

You have a password field with a length validator of `{ min: 4, max: 10 }` but want to display a warning telling the user their password is **hella** weak if its under 6 characters. 

Solution:

Warning validators! 

How it works:

- They should be declared like any other validator
- Act like an assert. If the conditions are unsatisfied, should return an error message (just like any validator)
- These validators will be filtered out of the "required" validators and all messages will be put in the `warnings` collection. 
- It should be up to the user when to display those warnings but they will not have any effect on the overall valid state of the attribute. 

Usage:

```js
password: {
  description: 'Password',
  validators: [
    validator('length', {
      min: 4,
      max: 10
    }),
    validator('length', {
      isWarning: true,
      min: 6,
      message: 'What kind of weak password is that?'
    })
  ]
}
```

screen shot 2016-06-23 at 4 19 10 pm

@rwjblue @stefanpenner @blimmer I would love to hear some input from your guys. Questions, concerns, whatever

### DIFF
--- a/addon/validations/factory.js
+++ b/addon/validations/factory.js
@@ -396,6 +396,14 @@ function createTopLevelPropsMixin(validatableAttrs) {
       return get(this, 'messages.0');
     })).readOnly(),
 
+    warnings: computed(...validatableAttrs.map(attr => `attrs.${attr}.@each.warning`), function () {
+      return emberArray(flatten(validatableAttrs.map(attr => get(this, `attrs.${attr}.warning`)))).compact();
+    }).readOnly(),
+
+    warning: computed('warning.[]', cycleBreaker(function () {
+      return get(this, 'warning.0');
+    })).readOnly(),
+
     errors: computed(...validatableAttrs.map(attr => `attrs.${attr}.@each.errors`), function () {
       return emberArray(flatten(validatableAttrs.map(attr => get(this, `attrs.${attr}.errors`)))).compact();
     }).readOnly(),
--- a/addon/validations/factory.js
+++ b/addon/validations/factory.js
@@ -393,15 +393,23 @@ function createTopLevelPropsMixin(validatableAttrs) {
     }).readOnly(),
 
     message: computed('messages.[]', cycleBreaker(function () {
-      return get(this, 'messages.0');
+      return get(this, 'messages.firstObject');
     })).readOnly(),
 
-    warnings: computed(...validatableAttrs.map(attr => `attrs.${attr}.@each.warning`), function () {
-      return emberArray(flatten(validatableAttrs.map(attr => get(this, `attrs.${attr}.warning`)))).compact();
+    warningMessages: computed(...validatableAttrs.map(attr => `attrs.${attr}.warningMessages`), function () {
+      return emberArray(flatten(validatableAttrs.map(attr => get(this, `attrs.${attr}.warningMessages`)))).compact();
+    }).readOnly(),
+
+    warningMessage: computed('warningMessages.[]', cycleBreaker(function () {
+      return get(this, 'warningMessages.firstObject');
+    })).readOnly(),
+
+    warnings: computed(...validatableAttrs.map(attr => `attrs.${attr}.@each.warnings`), function () {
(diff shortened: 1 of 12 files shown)


offirgolan:
Questions that need answering:

1. Is warning the right property to have? Maybe `required` or `severity`?
2. Which attributes in `result-collection.js` should be dependent on required validations or on all validations?

addon/validations/factory.js
```diff
@@ -396,6 +396,14 @@ function createTopLevelPropsMixin(validatableAttrs) {
       return get(this, 'messages.0');
     })).readOnly(),
 
+    warnings: computed(...validatableAttrs.map(attr => `attrs.${attr}.@each.warning`), function () {
+      return emberArray(flatten(validatableAttrs.map(attr => get(this, `attrs.${attr}.warning`)))).compact();
+    }).readOnly(),
+
+    warning: computed('warning.[]', cycleBreaker(function () {
+      return get(this, 'warning.0');
```
blimmer: is `.0` the right thing to use here? I've always used `firstObject`.

tests/dummy/app/templates/components/validated-input.hbs
```diff
@@ -5,10 +5,16 @@
...
-              {{v-get model valuePath 'message'}}
+              {{get (v-get model valuePath 'error') 'message'}}
           </div>
       {{/if}}
+
+      {{#if showWarningMessage}}
+        <div class="warning">
+            {{get (v-get model valuePath 'warning') 'message'}}
+        </div>
+      {{/if}}
```
blimmer: I wonder if we should expose all the messages for an attribute somehow. Since this is new functionality, I wonder if `messages` should return all messages, including warnings. I could see wanting to do something like this for my flavor of validated input:
```handlebars
{{#each (v-get model valuePath 'messages') as |msg|}}
  <div class="message {{if msg.isWarning 'warning'}}">
     {{msg}}
  </div>
{{/each}}
```

That way I don't have to have two blocks and reduce boilerplate templating. What do you think?

addon/validations/factory.js
offirgolan: This is the same code that is the `errors` and `error` CPs. I believe I tried firstObject but it errored on me. Ill give it another shot though.

tests/dummy/app/templates/components/validated-input.hbs
offirgolan: My theory was:

1. Not to break current API since there could be other addons that are dependent that the messages array will return a collection of strings that are required.

2. I wanted to separate out warning and errors because errors are whats important. Warnings are optional and should affect the state of the model in any way. 

We could release 3.0 with API changes and think of a better way to handle this

blimmer: Ah, of course - I blanked that messages are strings on the current API. Definitely don't want to break that.

Though warnings are optional, it seems like anytime someone took the time to write a validation for a warning, they'd want to expose it to the user, likely just with another class to style differently.

addon/validations/factory.js
blimmer: No problem - definitely not required, I just haven't seen this syntax before.

stefanpenner: firstObject ideally

addon/validations/result-collection.js
```diff
@@ -195,6 +195,41 @@ export default Ember.Object.extend({
...
+   * // Example
+   * get(user, 'validations.warning')
+   * get(user, 'validations.attrs.username.warning')
+   * ```
+   *
+   * @property warning
+   * @readOnly
+   * @type {Ember.ComputedProperty | Error}
+   */
+  warning: computed('warnings.[]', cycleBreaker(function () {
```
stefanpenner: `firstObject`

stefanpenner:
thoughts on `warning: true` -> `isWarning: true`

offirgolan:
@stefanpenner 👍

PR #229, opened 2016-06-25
Allow computed properties in validator options

xcambar:
It is currently not possible in ember-cp-validations to use computed properties as options for the validators (eg, to compute what should be the minLength of the field).

Simply by using `Ember.get(options, 'propertyName')` instead of `options.propertyName` in the validators, the implementor is allowed to benefit of computed properties.

I hope you'll like it!

### DIFF
--- a/addon/validators/alias.js
+++ b/addon/validators/alias.js
@@ -77,7 +77,7 @@ const Alias = Base.extend({
 
 Alias.reopenClass({
   getDependentsFor(attribute, options) {
-    const alias = typeof options === 'string' ? options : options.alias;
+    const alias = typeof options === 'string' ? options : get(options, 'alias');
     return [ `${alias}.messages.[]`, `${alias}.isTruelyValid` ];
   }
 });
--- a/addon/validators/base.js
+++ b/addon/validators/base.js
@@ -10,6 +10,7 @@ import { unwrapString } from 'ember-cp-validations/utils/utils';
 
 const {
   get,
+  getWithDefault,
   set,
   isNone
 } = Ember;
@@ -115,7 +116,7 @@ const Base = Ember.Object.extend({
 
     // Overwrite the validator's value method if it exists in the options and remove it since
     // there is no need for it to be passed around
-    this.value = builtOptions.value || this.value;
+    this.value = getWithDefault(builtOptions, 'value', this.value);
     delete builtOptions.value;
 
     return builtOptions;
@@ -224,7 +225,7 @@ const Base = Ember.Object.extend({
    */
   createErrorMessage(type, value, options = {}) {
     const messages = this.get('errorMessages');
-    let message = unwrapString(options.message);
+    let message = unwrapString(get(options, 'message'));
 
     options.description = messages.getDescriptionFor(get(this, 'attribute'), options);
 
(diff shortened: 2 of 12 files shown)


blimmer:
This seems reasonable but I definitely want @offirgolan to have a look at this. @xcambar is there any test coverage you could add for the functionality?

offirgolan:
@xcambar thank you for taking the time to put this together! I have a couple of concerns though:

- Is this really needed when all options can be functions which have access to all information needed to compute the necessary value? 

```js
validator('length', {
  min(model, attribute) {
    return model.get('meta.minLength');
  }
);
```

- The options object is just a pojo which doesnt have any support for computed properties. If it does, then that is because of a bug in Ember. 

```js
var options = {
  foo: 'isFoo',
  bar: computed.alias('foo')
}
```

The above wont work unless you do something like `options = Ember.Object.create(options)`

- If this was to actually work, the current context of the CP would just be other options. You wont have access to any public validator properties such as `model`, `attribute`, etc.

xcambar:
> Is this really needed when all options can be functions which have access to all information needed to compute the necessary value?

This is correct, but isn't the "option as function" variation a computed property in disguise?

> The options object is just a pojo which doesnt have any support for computed properties.

Correct again, but this PR allows POJOs **and more**. 

Here's my use case: I have a set of validation models (yes, as in Ember Data models). Those models have relationships and (computed) properties that depend on the relationships between them. As an example, "maiden_name" is only required when `gender === "woman"` and `married === true`. This is extremely easy to handle these use cases with computed properties (the framework does that for us).

Instead of recomputing the array of validators whenever the conditions apply, it's much easier to pass objects with CPs to the validator and have the correct values applied during validation.

offirgolan:
Okay so I've been thinking about this for a while and I think it can be
possible. It needs a little more internal work than what you have here but
I think we can make it work and have it totally replace the functional
options approach.

When I get home, I'll create a new branch and if you can create this exact
pr for that branch, I'll work on the rest.

I'll post on her when the branch has been created.

On Saturday, June 25, 2016, Xavier Cambar <notifications@github.com> wrote:

> Is this really needed when all options can be functions which have access
> to all information needed to compute the necessary value?
>
> This is correct, but isn't the "option as function" variation a computed
> property in disguise?
>
> The options object is just a pojo which doesnt have any support for
> computed properties.
>
> Correct again, but this PR allows POJOs *and more*.
>
> Here's my use case: I have a set of validation models (yes, as in Ember
> Data models). Those models have relationships and (computed) properties
> that depend on the relationships between them. As an example, "maiden_name"
> is only required when gender === "woman" and married === true. This is
> extremely easy to handle these use cases with computed properties (the
> framework does that for us).
>
> Instead of recomputing the array of validators whenever the conditions
> apply, it's much easier to pass objects with CPs to the validator and have
> the correct values applied during validation.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/offirgolan/ember-cp-validations/pull/229#issuecomment-228576675>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAjJwrQ-5U8eI2oj6dYDBoh488GwheBwks5qPcGdgaJpZM4I-ZNL>
> .
>


-- 
Thanks,
Offir Golan

offirgolan:
@xcambar I've created a branch `computed-options` please reopen this PR targeting that.

PR #231, opened 2016-06-26
Computed options

xcambar:
A follow up on #230

PR #232, opened 2016-06-27
[Feature] Computed Options 🔥

offirgolan:
Replace `Options as Functions` with CPs

```js
username: {
    dependentKeys: ['model.foo'],
    description: computed.alias('model.usernameDesc'),
    validators: [
      validator('presence', true),
      validator('length', {
        min: computed.alias('model.minLength')
      })
    ]
  }
```

Features:

1. Any property can now be a CP with access to the `model` and `attribute`
2. Dependent keys are __extracted__ from the CPs and added to the top level CP (`validations.attrs.username`)
3. Options that must be constantly recomputed, you can just use `volatile` 
4. `dependentKeys: ['foo']` --> `dependentKeys: ['model.foo']` to match CP dependentKey declaration

Special thanks to @xcambar for first introducing this idea to me. 

Open for review: @stefanpenner @rwjblue @blimmer

### DIFF
--- a/addon/validations/factory.js
+++ b/addon/validations/factory.js
@@ -342,7 +342,7 @@ function createCPValidationFor(attribute, validations, owner) {
     const validators = !isNone(model) ? getValidatorsFor(attribute, model) : [];
 
     const validationResults = validators.map(validator => {
-      const options = validator.processOptions();
+      const options = get(validator, 'options').copy();
       const debounce = getWithDefault(options, 'debounce', 0);
       const disabled = getWithDefault(options, 'disabled', false);
       let value;
@@ -441,8 +441,22 @@ function getCPDependentKeysFor(attribute, validations, owner) {
       getWithDefault(validation, 'globalOptions.dependentKeys', [])
     );
 
+    // Process dependentKeys from the CPs
+    const cpDependents = Object.keys(options).reduce((arr, key) => {
+      let option = options[key];
+
+      if(typeof option === 'object' && get(option, 'isDescriptor')) {
+        return arr.concat(get(option, '_dependentKeys').map(d => {
+          return d.split('.')[0] === 'model' ? '_' + d : d;
+        }));
+      }
+
+      return arr;
+    }, []);
+
     return baseDependents.concat(
       dependents,
+      cpDependents,
       specifiedDependents.map(d => `_model.${d}`)
     );
   });
@@ -465,7 +479,7 @@ function getCPDependentKeysFor(attribute, validations, owner) {
  * @param  {Function} resolve
  */
 function debouncedValidate(validator, model, attribute, resolve) {
-  const options = validator.processOptions();
+  const options = get(validator, 'options').copy();
   const value = validator.getValue();
 
   resolve(validator.validate(value, options, model, attribute));
--- a/addon/validations/factory.js
+++ b/addon/validations/factory.js
@@ -434,7 +434,6 @@ function getCPDependentKeysFor(attribute, validations, owner) {
     const Validator = type === 'function' ? BaseValidator : lookupValidator(owner, type);
     const baseDependents = BaseValidator.getDependentsFor(attribute, options) || [];
     const dependents = Validator.getDependentsFor(attribute, options) || [];
-    let cpDependents = [];
 
     const specifiedDependents = [].concat(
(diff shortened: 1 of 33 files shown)


addon/validators/base.js
```diff
@@ -118,32 +111,23 @@ const Base = Ember.Object.extend({
...
+          return obj;
+        }, {}));
       }
     });
 
-    // Cache the options so it can be accessed without triggering this method again
-    set(this, '_cachedOptions', assign({}, options));
-
-    return options;
+    return OptionsClass.create();
```
offirgolan: I attempted to do the following:

`return Object.freeze(OptionsClass.create());`

But it only worked in Ember 2.3+. Older version trigger a set on the object after create which throws an error. Maybe we can find another solution to this?

addon/validations/factory.js
```diff
@@ -454,6 +462,33 @@ function getCPDependentKeysFor(attribute, validations, owner) {
...
+ * @param  {Object} options
+ * @return {Array}  dependentKeys
+ */
+function extractOptionsDependentKeys(options) {
+  if(options && typeof options === 'object') {
+    return Object.keys(options).reduce((arr, key) => {
+      let option = options[key];
+
+      if(typeof option === 'object' && option.isDescriptor) {
+        let dependentKeys = option._dependentKeys || [];
```
offirgolan: Will this work in previous and future version of Ember?

addon/validators/base.js
xcambar: You can go with `Ember.Freezable`, but it's not quite the same as `Object.freeze`.

Or just leave it as is and live by the `readOnly` CPs, which, IMO, is good enough.

addon/validations/factory.js
xcambar: I'm pretty sure the `dependentKeys` option (and henceforth, extraction) is a thing from the past with this PR. I need a bit more explanation on why it's still required. (I see where it's used, I think we can get rid of it)

xcambar:
Overall very good PR, I'm really looking forward to seeing it live!

I see the previous PRs (#229 and #231) have been put to good use, congrats!

addon/validators/base.js
offirgolan: Ember.Freezable is deprecated. I think leaving it as is  for now is okay, but the original options object should be immutable. Any changes needed to make should be done via a copy.

addon/validations/factory.js
offirgolan: There could def. be use-cases when you want your validation to be dependent on another model All im doing is extracting the dependent keys from the option’s dependents so you dont have to specify it twice. So lets say I want my username validator to recompute when `foo` changes, you would have to add `foo` to your dependentKeys since it wont be in any other specified option 

`validator('length', { min: 5, dependentKeys: ['foo'] })`

offirgolan:
Follow up question:

The option CP's dependentKeys are all prefixed with `model`

```js
validator('length', { min: computed.alias('model.minLength') })`
```

Should we force all custom dependentKeys to be prefixed with `model` to avoid confusion?

`dependentKeys: ['foo']` --> `dependentKeys: ['model.foo']`

stefanpenner:
thoughts on something like `dependentKeys: ['model.foo']` -> `validatedProprty(list, of, dk, { options })` ? My thought it is make the API more similar to `Ember.computed`. This is largely a brainstorm.

offirgolan:
@stefanpenner not sure that would make too much sense. `dependentKeys` is just an additional collection of dependents for the top level CP.

stefanpenner:
```
username: validedProperty('model.foo', {
    description: computed.alias('model.usernameDesc'),
    validators: [
      validator('presence', true),
      validator('length', {
        min: computed.alias('model.minLength')
      })
    ]
  }
```

offirgolan:
That would make sense only for that use case.

What about when the dependentKeys are as such

```js
username: {
    description: computed.alias('model.usernameDesc'),
    validators: [
      validator('presence', true),
      validator('length', {
        dependentKeys: ['model.foo'],
        min: computed.alias('model.minLength')
      })
    ]
  }
```

stefanpenner:
@offirgolan ah, i see, It seems like your existing design is actually more consistent.

offirgolan:
@stefanpenner agreed. 😺

PR #239, opened 2016-06-28
[Bug Fix] Use Require for Checking Ember Data

### LINKED ISSUE #224: Ember-Data is required to use the DS Error validator - Too Strict for ED 2.6+

offirgolan:
Closes #224

### DIFF
--- a/addon/utils/utils.js
+++ b/addon/utils/utils.js
@@ -4,9 +4,14 @@
  */
 
 import Ember from 'ember';
+import require from 'require';
 
 export function hasEmberData() {
-  return typeof self.DS !== 'undefined';
+  return require.has('ember-data');
+}
+
+export function getEmberData() {
+  return hasEmberData() ? require('ember-data')['default'] : undefined;
 }
 
 export function unwrapString(input) {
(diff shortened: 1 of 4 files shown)


addon/utils/utils.js
```diff
@@ -4,9 +4,14 @@
  */
 
 import Ember from 'ember';
+import require from 'require';
 
 export function hasEmberData() {
-  return typeof self.DS !== 'undefined';
+  return require.has('ember-data');
```
alexparker: I'm getting an error here `utils.js Uncaught TypeError: Cannot read property 'has' of undefined` is there a version of ember this is incompatible with? I'm on ember 2.5

offirgolan: Weird, this has been tested on Ember 1.11 to canary. What version of ember-cli are you on?

/cc @stefanpenner

PR #240, opened 2016-06-28
[Bug Fix] DS Error Validator + Nested Keys

### LINKED ISSUE #236: ds-errors validator does not work in component
ds-errors is great when embedded directly in the model, but when used in a component, e.g.

```js
import { buildValidations, validator } from 'ember-cp-validations';

  'model.email': [
    validator('ds-error')
  ]
});

export default Ember.Component.extend(Validations, {
  actions: {
    saveMe() {
      this.get('model').save();
    }
  }
});
```

I think the `validate` function never fires when a model is saved in this way.

offirgolan:
Resolves #236

### DIFF
--- a/addon/utils/utils.js
+++ b/addon/utils/utils.js
@@ -4,9 +4,14 @@
  */
 
 import Ember from 'ember';
+import require from 'require';
 
 export function hasEmberData() {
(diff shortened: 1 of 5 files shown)


blimmer:
lgtm!

PR #241, opened 2016-06-29
Fix blueprint warning.

### LINKED ISSUE #233: Error creating validator test with blueprint
```
ember g validator-test my-validator
version: 2.4.2
Cannot read property 'addonPackages' of undefined
TypeError: Cannot read property 'addonPackages' of undefined
    at Class.module.exports._filesPath (foo/ember2.4-bar/node_modules/ember-cp-validations/blueprints/validator-test/index.js:20:42)
    at Class.Blueprint.files (foo/ember2.4-bar/node_modules/ember-cli/lib/models/blueprint.js:317:24)
    at Class.Blueprint.supportsAddon (foo/ember2.4-bar/node_modules/ember-cli/lib/models/blueprint.js:808:15)
    at Class.module.exports.Task.extend.run (foo/ember2.4-bar/node_modules/ember-cli/lib/tasks/generate-from-blueprint.js:25:98)
    at Class.module.exports.Command.extend.run (foo/ember2.4-bar/node_modules/ember-cli/lib/commands/generate.js:65:17)
    at Class.<anonymous> (foo/ember2.4-bar/node_modules/ember-cli/lib/models/command.js:152:17)
    at lib$rsvp$$internal$$tryCatch (foo/ember2.4-bar/node_modules/ember-cli/node_modules/rsvp/dist/rsvp.js:1036:16)
    at lib$rsvp$$internal$$invokeCallback (foo/ember2.4-bar/node_modules/ember-cli/node_modules/rsvp/dist/rsvp.js:1048:17)
    at lib$rsvp$$internal$$publish (foo/ember2.4-bar/node_modules/ember-cli/node_modules/rsvp/dist/rsvp.js:1019:11)
    at lib$rsvp$asap$$flush (foo/ember2.4-bar/node_modules/ember-cli/node_modules/rsvp/dist/rsvp.js:1198:9)
```

@blimmer

blimmer:
Resolves #233.

### DIFF
--- a/blueprints/validator-test/index.js
+++ b/blueprints/validator-test/index.js
@@ -47,5 +47,10 @@ module.exports = {
 
   _legacySrcPathOverride: function(file) {
     return path.resolve(this._filesPath(), file);
-  }
+  },
+
+  // workaround https://github.com/ember-cli/ember-cli/issues/5481
+  supportsAddon: function() {
+    return false;
+  },
 };


blueprints/validator-test/index.js
```diff
@@ -47,5 +47,10 @@ module.exports = {
 
   _legacySrcPathOverride: function(file) {
     return path.resolve(this._filesPath(), file);
-  }
+  },
+
+  // workaround https://github.com/ember-cli/ember-cli/issues/5481
+  supportsAddon: function() {
+    return false;
```
offirgolan: @blimmer what does this actually do?

blimmer: This is what's actually throwing the error - it's defined [here](https://github.com/ember-cli/ember-cli/blob/d8ba7dbaf77df0e7fe2daf1d153a1a5a60620143/lib/models/blueprint.js#L763-L773). It is defined as this:

```javascript
return !!this.files().join().match(/__root__/);
```

The `__root__` token is used in blueprints to allow being defined in either `app` or `addon` depending on context. However, this will always (and should always) return false for the test generator because, no matter the context, tests are always placed in `tests/` in an addon or app. By allowing this to short-circuit, we skip over the call to `files()` in the default definition which throws the error.

offirgolan: 👍 makes sense. Thanks for figuring out the workaround!

PR #242, opened 2016-06-29
Use ember-try version compatibilty for better coverage

blimmer:
Since we're missing post 2.3 in the current ember try.


blimmer:
Will this still work with codecoverage?

offirgolan:
@blimmer Im not really sure how `versionCompatibility` will work when we have different dependencies for different versions of Ember (i.e. ember-data and ember-get-helper). I know with the latest ember-try we can have multiple configs so maybe we can split ours up to:

1. Ember >= 1.11 < 2.0
2. Ember 2.0 - 2.6.1
3. Ember beta, canary, and release

See [here](https://github.com/ember-cli/ember-try#ember-tryconfig)

blimmer:
Hmm didn't realize the inner complexity. I actually don't have a ton of time to look at the moment. I'll close and look into it more if I have some time.

offirgolan:
👍

PR #245, opened 2016-07-02
Utilize __root__ in validator blueprint.

blimmer:
This allows placing the validator in addon/ for addons and app/ for app.


blimmer:
Addon:
```bash
› ember g validator foo-bar
installing validator
  create addon/validators/foo-bar.js
installing validator-test
  create tests/unit/validators/foo-bar-test.js
installing addon-import
  create app/validators/foo-bar.js
```

App:
```bash
› ember g validator foo-bar
installing validator
  create app/validators/foo-bar.js
installing validator-test
  create tests/unit/validators/foo-bar-test.js
```

PR #249, opened 2016-07-05
fixed email regex of format validator

simonihmig:
Hi Offir,
I stumbled upon the fact that the email type format validator did not catch an obviously invalid address like "mail@domain". There were some other edge cases where it failed as well (that is it failed in not failing ;)). I covered those cases in the updated test, and updated the email regex (taken from http://www.regular-expressions.info/email.html)


offirgolan:
@simonihmig thanks for taking the time to modify the email regex. I think the changes made are great, but Im not too sure about removing uppercase letters. In most forms, they are accepted but are in the end, lowercased by the mail server.

simonihmig:
That's why I added the `i` flag at the end of the regex! ;)
Just added another test case for uppercase letters, to make sure it works. But this was already covered by the regex...

offirgolan:
@simonihmig 💯 💯 💯 

I dont know how I missed that! Thanks for clarifying and for the extra test case 😸

GreatWizard:
I just notice this PR and I would like to say that `mail@domain` is real a valid email adress. On a local network you can easily have `user@localdomain`. It's a breaking changes, we need to have a warning in the documentation on this specific point.

indr:
Thank you @simonihmig for providing this new regex. I was looking for that because my application needs email addresses with top-level domains. But in general an email address like `mail@localhost` is valid, too.

Also, this pull request may introduce breaking changes for some people. My suggestion therefore is to add another validator format `emailTld`, or make the tld regex the default and add `emailNonTld` for these folks. Anyway, it's not difficult to reopen the class and add another format to the `regularExpressions` key.

/cc @offirgolan

simonihmig:
To be honest I am no expert regarding all the affected RFCs out there that affect the validity of email addresses. It even seems to be controversial between RFCs (http://stackoverflow.com/questions/83723/allow-userexample-or-userlocalhost-in-email-validation). 

I still think requiring a tld is a reasonable default, as a @localhost domain would probably only work on some internal intranet type of application.

My suggestions:

* keep it as it is. We should probably add some note in the docs about this. One could easily define a custom email validator by not using a `type` of "email", but using the `regex` property directly for a custom regex
* we could introduce a `allowNonTld` flag, which is false by default. Something like this:
```js
validator('format', {
  type: 'email',
  allowNonTld: true
})
``` 

What's your point on this, @offirgolan ? I won't be able to tackle this in the near time however, as I will be on vacation soon...

simonihmig:
It seems I just pinged a github user called "localhost" in the comment above. Ha ha! :) Sorry man, did not want to bother you! ;)

indr:
I would also leave it as it is. The majority of ember apps run in the Internet. I like that `allowNonTld` flag, but default `false`, because that's what devs expect. The others should read Changelogs ;)
Thanks @simonihmig !

offirgolan:
@indr I think this is a valid concern and reason to introduce a `allowNonTld` option to the mix. Can you please open this up in a new issue so we can easily track this? Im currently in the process of migrating a lot of the validators into another addon so I dont want to forgot about this issue.

@simonihmig no worries on not being able to work on this. I can handle this one 😏

PR #252, opened 2016-07-14
[BUGFIX] Fix require module

### LINKED ISSUE #250: Error after upgrade to 3
As soon as I import I get an error coming from ember-cp-validations/utils/utils the hasEmberData method.

```
import {
  validator,
  buildValidations
}
from 'ember-cp-validations';
```
```
[Error] TypeError: undefined is not an object (evaluating '_require['default'].has')
	hasEmberData (vendor.js:86444)
	getEmberData (vendor.js:86448)
	....
```

offirgolan:
Resolves #250.

@basz @alexparker can you checkout this branch and see if that fixes your problem? Im not able to recreate the issue from my end but I think this should do it.

### DIFF
--- a/addon/utils/utils.js
+++ b/addon/utils/utils.js
@@ -4,14 +4,9 @@
  */
 
 import Ember from 'ember';
-import require from 'require';
 
-export function hasEmberData() {
-  return require.has('ember-data');
-}
-
-export function getEmberData() {
-  return hasEmberData() ? require('ember-data')['default'] : undefined;
+export function requireModule(module) {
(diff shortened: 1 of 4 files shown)


basz:
Yes, this solve my issue. Seems to work just fine - thank you!

alexparker:
👍 works for me

offirgolan:
Shweet! Thanks for confirming. Ill release a new 3.0 beta with this PR

jasonmit:
@offirgolan this relies on the newer API of loader.js and older versions (<= 4.0.11) do not have the `has` method.  Perhaps fall back to the previous lookup behavior when `has` is not available?

PR #266, opened 2016-07-26
Check for null in extractOptionsDependentKeys

xcambar:
Changes proposed:

 - Adds a check for null in `extractOptionsDependentKeys` (because `typeof null === "object"`)


offirgolan:
👍 good catch. Thank you!

PR #270, opened 2016-07-26
Fix null value for the message

GreatWizard:
Changes proposed:

 - on `extractOptionsDependentKeys`: `typeof null === "object"`

Tasks:

- [x] Added test case(s): The default message is available!


offirgolan:
@GreatWizard Thanks for adding this test case 👍

PR #312, opened 2016-08-11
[BUGFIX] Deleted DS.Model records should be suppressed

### LINKED ISSUE #304: Deleted records are still validated
### Environment

Ember Version: 2.6.3
Ember CLI Version: 2.6
Ember CP Validations Version: 2.9.3

### Steps to Reproduce

I was doing some work on our app and was curious if deleted, but not saved records would still validate, and they do. I believe the has-many validator should be updated to ignore deleted but not saved to the server records.

I wrote a test based on my previous ones to show this functionality. 

Also, made me think about the presence validator and belongs-to. If a record is deleted but not saved, and presence: true is there, is that record valid? 

```js
test("order with invalid question shows valid if invalid question is deleted in reverse order", function(assert) {
  assert.expect(8);
  var order = this.subject({ id: 1, source: 'external' });

  var orderLine, orderSelection, orderSelectionQuestion;

  let store = this.store();
  Ember.run(() => {
    let fakeSave = function(model) {
      model.get('_internalModel').adapterWillCommit();
      model.get('_internalModel').adapterDidCommit();
    };

    orderLine = store.createRecord('order-line', { id: 1, order: order, type: "item" });
    orderSelection = store.createRecord('order-selection', { id: 1, order: order, line: orderLine, quantity: 1 });
    orderSelectionQuestion = store.createRecord('order-selection-question', { id: 1, order: order, selection: orderSelection, text: 'answer' });
    orderSelectionQuestion2 = store.createRecord('order-selection-question', { id: 2, order: order, selection: orderSelection });

    fakeSave(order);
    fakeSave(orderLine);
    fakeSave(orderSelection);
    fakeSave(orderSelectionQuestion);
    fakeSave(orderSelectionQuestion2);
  });

  assert.equal(order.get('lines.length'), 1, 'Order has 1 Order Line');
  assert.equal(orderLine.get('selections.length'), 1, 'Order Line has 1 Order Selection');
  assert.equal(orderSelection.get('questions.length'), 2, 'Order Selection has 2 Order Selection Question');

  let done = assert.async();

  orderSelectionQuestion2.validate().then(({ validations }) => {
    assert.equal(validations.get('isValid'), false, 'Order Selection Question 2 should be invalid');
    return orderSelectionQuestion.validate();
  }).then(({ validations }) => {
    assert.equal(validations.get('isValid'), true, 'Order Selection Question should be valid');
    return orderSelection.validate();
  }).then(({ validations }) => {
    assert.equal(validations.get('isValid'), false, 'Order Selection should be invalid because of Order Selection Question 2');
    orderSelectionQuestion2.deleteRecord();
    return orderSelection.validate();
  }).then(({ validations }) => {
    assert.equal(orderSelection.get('questions.length'), 2, 'Order Selection has 2 Order Selection Question');
    assert.equal(validations.get('isValid'), true, 'Order Selection should be valid because invalid Order Selection Question 2 was marked deleted');
    done();
  });
});
```

offirgolan:
Resolves #304.

Changes proposed:

 - If a DS.Model record is in a deleted state, validations on that record should be suppressed
 - Create CPs per class and not per instance 

Open Questions:

1. I believe this solves the first part of the issue. The question now lies as to how do external validators deal with a deleted record. @aaronbhansen mentioned only adding support to this in the `presence` validator which would be simple but about the `length` validator (i.e. a has-many relationship must have at least 1 record in it)? I dont think such logic should be dealt with by the validator.

2. ~~Should this logic be applied for an `isDestroyed` object? How do we handle a has-many relationship or a belongs-to relationship with an Ember.Object that has been destroyed?~~

/cc @aaronbhansen @stefanpenner @rwjblue 

Tasks:

- [x] Added test case(s)
- [x] Fix ember dependent test failures in build
- [x] Updated documentation


addon/validations/factory.js
```diff
@@ -305,36 +306,52 @@ function createValidationsClass(inheritedValidationsClass, validations, owner) {
...
     init() {
       this._super(...arguments);
-      const model = this.get('_model');
 
-      // Create the CPs
+      const _model = this.get('_model');
+
+      /*
+        Assign the CPs to this object
+        TODO: This runs multiple nested defineProperty calls PER INSTANCE which
```
offirgolan: @stefanpenner is there any way around this?

aaronbhansen:
Looking at all the validators, I think there are only four that could apply to Ember Data models or an object that could be destroyed. Here are my thoughts. 

* belongs-to
  * When deleted, should ignore any validations on the belongs-to
* has-many
  * Ignores any child records validations that are deleted.
* length
  * Shouldn't include in the collection length any deleted records. 
* presence
  * If a belongs-to record exists, but is deleted, its not considered present.

The ones that don't seem to apply are:

`collection, confirmation, date, dependent, exclusion, format, inclusion`

Anyway, those are my thoughts on how the validations might apply to deleted or destroyed objects.

offirgolan:
@aaronbhansen I was able to come up with a much more scalable solution that doesnt involve any extra logic on the validator side. This way, even if new validators get added later on, they wont have to include any DS.Model deletion logic. 

If you have the time, it would be awesome if you can use this branch in your app to test all this out 😺

addon/utils/utils.js
```diff
@@ -7,18 +7,44 @@ import Ember from 'ember';
...
+export function unwrapString(s) {
+  if (s && s instanceof Ember.Handlebars.SafeString) {
+    return s.toString();
+  }
+
+  return s;
+}
+
+export function unwrapProxy(o) {
+  if (o && (o instanceof Ember.ObjectProxy || o instanceof Ember.ArrayProxy)) {
```
aaronbhansen: You could simplify this to cover anything using the mixin if you wanted. 

```
if (o && (Ember.PromiseProxyMixin.detect(o))
```

Covers a bigger range of promises with a content. Just a thought. I use it in from https://github.com/unchartedcode/ember-promise-tools/blob/master/addon/utils/get-promise-content.js

aaronbhansen:
Thats great @offirgolan I'll try to check it out over the weekend. 💯

addon/utils/utils.js
offirgolan: @aaronbhansen `Ember.Mixin.detect` is private API and can change at any time without warning 😞 .

aaronbhansen:
Changes look good. I like how you refactored the code to make this work. We should have our refactoring down hopefully end of this week then I can more fully test it.

offirgolan:
@aaronbhansen that would be great thank you. Seems like test are failing in different env. since record.deleteRecord is not consistent across ember-data versions. Im gonna have to do a dummy delete instead. Hopefully once I get them to pass, I can get this merged and released the next beta 😸

PR #323, opened 2016-08-17
fixing input red borders in dummy app

MarcoUmpierrez:
Resolves # .
In the component `validated-input`, the property `showErrorClass` was checking an undefined property (`showMessage`) which always makes it false and the component never got the `has-error` class, preventing the border of the input to turn red.

Changes proposed:
 - Changed `showMessage` for `showErrorMessage` in [#L38](https://github.com/offirgolan/ember-cp-validations/blob/master/tests/dummy/app/components/validated-input.js#L38) of the component `validated-input`.


offirgolan:
Oh man I didnt even see that! Thank you for making this PR 😺 I guess I missed it when I split `showMessage` to `showErrorMessage` and `showWarningMessage`.

PR #330, opened 2016-08-27
Add option allowNonTld for email format validator

### LINKED ISSUE #329: Fixed email format validator with opt-out for top-level domain validation
@simonihmig changed the email regex of the format validator to include the presence of a top-level domain with PR #249 :+1: 

This will introduce a breaking change for people who expect email addresses like 'mail@localhost' to be valid. In the comments to the PR was this decision made:

- [x] Email format validator validates only if a top-level domain is present (PR #249)
- [ ] Introduce option `allowNonTld` with default value `false` (PR #330)
- [ ] Mention this breaking change together with the new option in the changelog

/cc @offirgolan

indr:
Resolves #329.

Tasks:

- [x] Added test case(s)
- [x] Updated documentation

### DIFF
--- a/addon/validators/format.js
+++ b/addon/validators/format.js
@@ -60,7 +60,7 @@ const {
 export default Base.extend({
   regularExpressions: {
     email: /^[a-z0-9!#$%&'*+\/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+\/=?^_`{|}~-]+)*@(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?$/i,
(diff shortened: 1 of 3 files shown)


addon/validators/format.js
```diff
@@ -76,7 +79,12 @@ export default Base.extend({
     const { regex, type } = options;
 
     if (type && !isNone(regularExpressions[type]) && isNone(regex)) {
-      options.regex = regularExpressions[type];
+      if (type === 'email' && options.allowNonTld) {
+        options.regex = regularExpressions['emailNonTld'];
```
offirgolan: As `emailNonTld` is a literal string, please replace it with `options.regex = regularExpressions.emailNonTld`

addon/validators/format.js
```diff
@@ -76,7 +79,12 @@ export default Base.extend({
...
+      }
+      else {
```
offirgolan: Please use 

```js
if(condition) {

} else {

}
```

offirgolan:
@indr thank you for taking the time in putting this PR together! This is fantastic 😸. I've added a couple styling comments. Once those are resolved I'll go ahead and merge this PR.

indr:
Wow, this is amazing, I just pushed and the PR gets updated :smile: 

@offirgolan My pleasure. Thanks for your comments. Will adapt this style for my projects.

offirgolan:
@indr looks great! Thanks again for your contribution. Ill make sure to note this as well as the earlier regression in the next release.

PR #338, opened 2016-09-11
Add validator type to error messages

kepek:
Changes proposed:

 - Add validator type to error messages. It can be helpful if you're checking if there's a specific type of error in the collection.

Tasks:

- [x] Added test case(s)
- [x] Updated documentation

```js
model.get('validations.errors').filterBy('type', 'unique-email');

// [ { type: 'presence',
//     message: 'This field can\'t be blank',
//     attribute: 'lastName' } ]
```

```js
model.get('validations.errors');

// [ { type: 'length',
//     message: 'This field is too short (minimum is 5 characters)',
//     attribute: 'firstName' },
//   { type: 'presence',
//     message: 'This field can\'t be blank',
//     attribute: 'lastName' } ]
```

```js
Validations = buildValidations({
  firstName: [
    validator('presence', true),
    validator('length', {min: 5, max: 35})
  ],
  lastName: [
    validator('presence', true),
    validator('length', {min: 5, max: 35})
  ]
});
```

```js
Ember.Object.extend(Validations), {
  firstName: 'Foo',
  lastName: null
});
```

### DIFF
--- a/addon/validations/internal-result-object.js
+++ b/addon/validations/internal-result-object.js
@@ -25,6 +25,7 @@ export default Ember.Object.extend({
   model: null,
   isValid: true,
   isValidating: false,
+  type: null,
   message: null,
   attribute: '',
 
@@ -72,9 +73,10 @@ export default Ember.Object.extend({
     return makeArray(get(this, 'message'));
   }),
 
-  error: computed('message', 'isInvalid', 'attribute', function () {
+  error: computed('isInvalid', 'type', 'message', 'attribute', function () {
     if (get(this, 'isInvalid')) {
       return ValidationError.create({
+        type: get(this, 'type'),
         message: get(this, 'message'),
         attribute: get(this, 'attribute')
       });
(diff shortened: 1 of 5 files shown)


addon/validations/internal-result-object.js
```diff
@@ -25,6 +25,7 @@ export default Ember.Object.extend({
   model: null,
   isValid: true,
   isValidating: false,
+  type: null,
```
offirgolan: `type: computed.readOnly('_validator._type')`

addon/validations/result.js
```diff
@@ -159,6 +159,7 @@ const Result = Ember.Object.extend({
    */
   _validations: computed('model', 'attribute', '_promise', '_validator', function () {
     return InternalResultObject.extend({
+      type: computed.readOnly('_validator._type'),
```
offirgolan: This can be removed.

kepek: Done

addon/validations/internal-result-object.js
kepek: True! Done!

kepek:
Not sure why Travis is complaining 😼  On my local env all seems to be fine: `Scenario ember-1.12: SUCCESS`

offirgolan:
@kepek no worries it just hangs from time to time. All good now! 

Thanks again for your contribution! 😺

PR #339, opened 2016-09-11
Allow for requirejs.has to not be available

jasonmit:
We spoke about this on slack a week back, older versions of the loader do not support the `has` API.

### DIFF
--- a/addon/utils/utils.js
+++ b/addon/utils/utils.js
@@ -15,7 +15,14 @@ const {
 } = Ember;
 
 export function requireModule(module) {
-  return self.requirejs.has(module) ? self.require(module).default : undefined;
+  const rjs = self.requirejs;
+
+  if (
+    (rjs.has && rjs.has(module)) ||
+    (!rjs.has && (rjs.entries[module] || rjs.entries[module + '/index']))
+  ) {
+    return self.require(module).default;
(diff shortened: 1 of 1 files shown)


addon/utils/utils.js
```diff
@@ -15,7 +15,14 @@ const {
...
 
 export function requireModule(module) {
-  return self.requirejs.has(module) ? self.require(module).default : undefined;
+  const rjs = self.requirejs;
+
+  if (
+    (rjs.has && rjs.has(module)) ||
+    (rjs.entries[module] || rjs.entries[module + '/index'])
+  ) {
+    return self.require(module).default;
```
offirgolan: @jasonmit does this silently fail if the module does not exist?

jasonmit: no, but I'm checking if it exists before accessing it which is what the previous logic was

offirgolan: 👍

offirgolan:
@jasonmit thank you for putting this together!

offirgolan:
@jasonmit this has been released in `v3.0.0`. A patch release for 2.x has also been released with this change (`v2.9.6`)

jasonmit:
Thanks

PR #344, opened 2016-09-14
DEPRECATION: Ember.Handlebars.SafeString

kepek:
```DEPRECATION: Ember.Handlebars.SafeString is deprecated in favor of Ember.String.htmlSafe [deprecation id: ember-htmlbars.ember-handlebars-safestring] See http://emberjs.com/deprecations/v2.x#toc_use-ember-string-htmlsafe-over-ember-handlebars-safestring for more details.```

Changes proposed:
 - Add polyfill for `Ember.Handlebars.SafeString` to keep backwards compatibility.

Tasks:
- N/A

PR #349, opened 2016-09-16
DEPRECATION: Ember.Handlebars.SafeString

Dhaulagiri:
Changes proposed:

 - Backports the fix from #344 to the 2.x branch for users still on that

(unsure if my changes are what is causing CI to fail or not)


offirgolan:
👍 Ill make a new patch release with this. Thank you @Dhaulagiri

PR #358, opened 2016-09-25
[BUGFIX] Remove targetObject usage in dummy app

### LINKED ISSUE #355: Usage of `targetObject` in example app
I noticed that your example app at http://offirgolan.github.io/ember-cp-validations/ uses the private `targetObject` API to read `didValidate` off the `targetObject` (which happens to be the controller in this case, but isn't guaranteed to be).

This is going to be deprecated in 2.9 (related issue: https://github.com/emberjs/ember.js/issues/14168) and I think, it's fair to say, is considered a really bad practice. Components should be passed the data they need to do their job, not read data magically from the context they're evaluated inside of. Metaphorically, in JS, this would be akin to doing something like `arguments.caller.someLocalVariable`.

offirgolan:
Resolves #355.

Changes proposed:

 - Remove targetObject usage in dummy app

### DIFF
--- a/tests/dummy/app/components/validated-input.js
+++ b/tests/dummy/app/components/validated-input.js
@@ -21,9 +21,9 @@ export default Ember.Component.extend({
   placeholder: '',
   validation: null,
   showValidations: false,
+  didValidate: false,
 
   notValidating: computed.not('validation.isValidating').readOnly(),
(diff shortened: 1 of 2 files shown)


rwjblue:
:clap:

PR #380, opened 2016-11-02
[BUGFIX] When the path is empty, use `this` for createAttrsClass

jasonmit:
#Resolves regression in Ember 2.10.0-beta.*.

Currently lastObject is undefined and results in Ember.get(lastObject, ...) to throw with cannot get on undefined.

Changes proposed:

 - Ember.get no longer will return the first argument when you pass an empty string as the key/path
 - When you pass an empty string, return `this` like the original behavior

### DIFF
--- a/addon/validations/factory.js
+++ b/addon/validations/factory.js
@@ -303,7 +303,8 @@ function createAttrsClass(validatableAttributes, validationRules, owner) {
       validatableAttributes.forEach(attribute => {
         // Add a reference to the model in the deepest object
         const path = attribute.split('.');
-        const lastObject = get(this, path.slice(0, path.length - 1).join('.'));
+        const pathFragment = path.slice(0, path.length - 1).join('.');
+        const lastObject = isEmpty(pathFragment) ? this : get(this, pathFragment);
 
         if (isNone(get(lastObject, '_model'))) {
           set(lastObject, '_model', model);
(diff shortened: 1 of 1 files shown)


offirgolan:
@jasonmit thanks for putting this together. Has this change been documented at all? Should we open an issue on the ember side noting this regression?

jasonmit:
> Has this change been documented at all? Should we open an issue on the ember side noting this regression?

Unsure if intentional or if an empty string was ever intended to return `this` and it was a bug that people relied on.  In any case, I do plan to open one upstream if one does not exist (a quick search didn't yield anything).

offirgolan:
Okay great. I think this fix is pretty safe. Much better than relying on ember to return `this`. I also noticed that this is pushing to the 2.x branch. Is this regression present in 3.x?

jasonmit:
@offirgolan unsure about 3.x since we don't target that (yet).

offirgolan:
Okay I'll make a patch release with this for 2.x and then cherry pick it for 3.x.

jasonmit:
CI pipeline not printing anything, so unsure what is going on there.

offirgolan:
@jasonmit not sure why tests are failing. I'll merge and see whats going on here.

offirgolan:
@jasonmit made a patch release with this (v2.9.8).

jasonmit:
Thanks @offirgolan

PR #419, opened 2016-12-19
WIP: New error api

luxferresum:
Related: https://github.com/jasonmit/ember-i18n-cp-validations/issues/24


offirgolan:
@luxferresum I think this solution is too messy and very error prone. I really dont like the whole passing the owner around to different files and private classes. 

I need some time to think this over and hopefully find a more appropriate solution.

luxferresum:
@offirgolan sure! I know this is mostly a proof of concept at this point. Just ping me if I can help.

PR #428, opened 2017-01-05
Adding canceling of debounced validations

estshy:
Let's say I have async email validation which will fire every keydown. In that case I want to debounce it every several milliseconds. I would assume that debounce will abort all promises before any new debounced function will start. Unfortunately it's not the case. In Ember CP Validations deboucing works more like deferring the validation because it never cancels previous async operation.

### DIFF
--- a/addon/validations/factory.js
+++ b/addon/validations/factory.js
@@ -476,7 +476,9 @@ function generateValidationResultsFor(attribute, model, validators, validate, op
         let t = run.debounce(validator, resolve, debounce, false);
 
         if (!opts.disableDebounceCache) {
-          cache[guidFor(validator)] = t;
+          let guid = guidFor(validator);
+          Ember.run.cancel(cache[guid])
+          cache[guid] = t;
         }
       }).then(() => {
         return validate(validator, get(validator, 'options').copy());
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "ember-cli-inject-live-reload": "^1.4.1",
     "ember-cli-jshint": "^2.0.1",
     "ember-cli-less": "1.5.3",
-    "ember-cli-moment-shim": "3.0.0",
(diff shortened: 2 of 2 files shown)


offirgolan:
@estshy thats not what should be happening. 

From the Ember.run.debounce docs:

> Delay calling the target method until the debounce period has elapsed with no additional debounce calls. If debounce is called again before the specified time has elapsed, the timer is reset and the entire period must pass again before the target method is called.

This means that the validator shouldnt even be executed until the specified time has elapsed without the debounce method getting called again. The only thing debounce does is guarantee that your validations will only be run **once** after the specified time has passed without changes to its dependents. Ideally, there should be any promises to abort because only a single promise should exists per a successful debounce.

estshy:
@offirgolan You are partially right. Debounce should work out of the box provided that you will pass to it reference to the same function but we are passing a new instance each time validators are being build otherwise there is no way for ember to know whether one function is actually the same as the other one.

I will make a demo for you to demonstrate that async validators are actually not debounced properly soon.

taheilo:
+1. Would love if it would cancel previously queued validations (within debounce timer).

offirgolan:
@estshy good catch! This seems to be a regression from a previous version after I changed it to use `resolve` as the passed method. 

I think at this point a neater solution would be to just use an "in-house" debounce. 

```js
value = new Promise((resolve) => {
          let t = run.later(resolve, debounce);
  
          if (!opts.disableDebounceCache) {
           let guid = guidFor(validator);
           run.cancel(cache[guid])
           cache[guid] = t;
          }
        })
```

Thoughts? If you can test that out as well as add a test case that would be fantastic!

offirgolan:
Closing this PR as I addressed it in #434.

offirgolan:
Released the fix in `v3.1.5`.

PR #434, opened 2017-01-12
[BUGFIX] Debounce context not properly set

offirgolan:
Changes proposed:

 - Create a static `resolveDebounce` method to pass run.debounce

Tasks:

- [x] Added test case(s)
- [x] Updated documentation

### DIFF
--- a/addon/validations/factory.js
+++ b/addon/validations/factory.js
@@ -473,7 +473,7 @@ function generateValidationResultsFor(attribute, model, validators, validate, op
 
       // Return a promise and pass the resolve method to the debounce handler
       value = new Promise((resolve) => {
-        let t = run.debounce(validator, resolve, debounce, false);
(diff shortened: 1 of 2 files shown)


estshy:
@offirgolan Are you sure it's going to work correctly with two async validations on the same time? I think one might be debouncing the other one.

offirgolan:
I dont think they should interfere since the debounce context is the validator BUT I could be wrong. Can you test this out? If not I can test it out sometime this weekend.

offirgolan:
@estshy after some digging, it looks like I'm correct 😛 

https://github.com/ebryn/backburner.js/blob/master/lib/backburner/index.js#L662

offirgolan:
@estshy here is some more proof:

https://ember-twiddle.com/?openFiles=components.my-component.js%2Ctemplates.components.my-component.hbs

estshy:
@offirgolan Correct. That's nice, always good to learn something.

PR #442, opened 2017-01-24
[BUGFIX] Replace private and deprecated `_lookupFactory` with public & performant `factoryFor`

cibernox:
No API surface has changed. `_factoryFor` will be removed in 2.13.

### DIFF
--- a/addon/-private/factory-for.js
+++ b/addon/-private/factory-for.js
@@ -1,0 +1,14 @@
+/**
+  Utility to use `ApplicationInstance#factoryFor` if available (2.12+) and
+  fallback to private `_lookupFactory` otherwise
+*/
+export default function factoryFor(owner, type) {
+  let factory;
+  if (owner.factoryFor) {
+    let maybeFactory = owner.factoryFor(type);
+    factory = maybeFactory && maybeFactory.class;
+  } else {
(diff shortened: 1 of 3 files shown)


offirgolan:
@cibernox thank you for this. Can you move the code to a util file? Something like:

```js
// addon/utils/lookup-class.js
export default function lookupClass(owner, type) {
  let klass;
  if (owner.factoryFor) {
        let maybeKlass = owner.factoryFor(type);
        klass = maybeErrorMessages && maybeErrorMessages.class;
      } else {
        klass = owner._lookupFactory(type);
      }
return klass;
}
```

cibernox:
sure

offirgolan:
@cibernox thanks! Also, is there a need to fallback to `_lookupFactory`? Shouldnt the getOwner polyfill take care of that?

cibernox:
@offirgolan It is needed. `getOwner` gives you an applicationInstance, but `ApplicationInstance#factoryFor` was added in 2.12.

cibernox:
@offirgolan I added it in `-private`, because it's for internal usage only.

offirgolan:
@cibernox released with `v3.2.2`. Thanks again for the help 😸

PR #443, opened 2017-01-26
[BUGFIX] Change factoryFor to use factoryFor().create()

jasonmit:
Discussion: https://github.com/jasonmit/ember-i18n-cp-validations/issues/29

This will extend types, with the owner preserved, while still allowing us to access class properties.

Squash on merge.


addon/validations/factory.js
```diff
@@ -189,6 +189,14 @@ function normalizeOptions(validations = {}, globalOptions = {}) {
   });
 }
 
+function classProperty(value, property) {
```
jasonmit: BaseValidator requires this since it's imported and there for does not go through `factoryFor`.  This method knows how to deal with reading constructor props off types output from factoryFor and regular types.

addon/-private/factory-for.js
```diff
@@ -2,13 +2,22 @@
   Utility to use `ApplicationInstance#factoryFor` if available (2.12+) and
   fallback to private `_lookupFactory` otherwise
 */
+class LegacyFactoryFor {
```
jasonmit: Just mimics the same public API shape, not meant to be a polyfill, but instead just provides the same expected shape now with the factoryFor methods/properties we rely on downstream within addon/validations/factory.js

addon/-private/factory-for.js
```diff
@@ -2,13 +2,22 @@
...
+      return this.class.create(fullName, options);
+    }
+  }
+}
+
 export default function factoryFor(owner, type) {
-  let factory;
   if (owner.factoryFor) {
-    let maybeFactory = owner.factoryFor(type);
-    factory = maybeFactory && maybeFactory.class;
```
jasonmit: Source of the issue is returning .class here which does not preserve the owner when instances are instantiated from it. We need to create() from that object that factoryFor returns while still being able to access static properties before instantiation.

addon/-private/factory-for.js
```diff
@@ -2,13 +2,30 @@
   Utility to use `ApplicationInstance#factoryFor` if available (2.12+) and
   fallback to private `_lookupFactory` otherwise
 */
-export default function factoryFor(owner, type) {
-  let factory;
+class LegacyFactoryForWrapper {
```
jasonmit: Mocks the same public API that we rely on so we can consistently use class/create throughout the rest of the app.

addon/-private/factory-for.js
```diff
@@ -2,13 +2,30 @@
...
+    if (this._factory) {
+      return this._factory.create(fullName, options);
+    }
+  }
+}
+
+export default function factoryFor(owner, fullName) {
   if (owner.factoryFor) {
-    let maybeFactory = owner.factoryFor(type);
-    factory = maybeFactory && maybeFactory.class;
```
jasonmit: The bug stemmed from this since we end up trying to instantiate object from `maybeFactory.class` which do not carry forward the owner.  What we want is really to return `maybeFactory`.

offirgolan:
@jasonmit this is a sleek approach but im not sure if its too much? To make the solution and code readability simpler, would it not be easier to just assume the no class returned from the factory carries over the owner and force inject it on create instead via `owner.ownerInjection()`? 

May I'm just missing something here... 

@rwjblue have you thought of polyfilling `factoryFor` to avoid this kind of problem?

jasonmit:
Unsure what you mean by "assume the no class returned from the factory carries over the owner".

jasonmit:
Any attempt to return the owner from the class would introduce an unnecessary extend which would defeat the new factoryFor API.

cibernox:
I think that we might not understand the new factory for. I've asked for help in slack and linked the original issue. Let's see what experts say is the right way of having the owner injected. I suspect that this might be because of the `reopen`.  Perhaps it should be a `reopenClass`

addon/-private/factory-for.js
jasonmit: It is if you're not accessing props off the class, which is only available now via the class prop.  Instances should not be created from that class prop AFAIK and which led to the owner not being injected.

jasonmit:
Should then, while we wait, revert the factoryFor PR as there is problems with anyone relying on the owner (i.e., inject.service)

offirgolan:
> Unsure what you mean by "assume the no class returned from the factory carries over the owner".

@jasonmit what I meant is that we just force inject the owner on create no matter what the situation is since we have access to the owner regardless. That should simplify the problem for now until we get a better alternative. 

```js
factoryFor('validator:presence').create(owner.ownerInjection(), options);
```

Same thing for the Messages class. Thoughts?

cibernox:
For the record, @rwjblue explained me the `factoryFor` thing in https://github.com/jamesarosen/ember-i18n/pull/431
Apparently it returns some sort of proxy that has a `create` method that already setups injections, so no need of returning the class.

Revert the change if you want, but I'm going to work on this in a moment. I want to create a failing test case first.

offirgolan:
Im not at my computer right now so I will look into this more in a few hours. Im curious to see if the returned object from factoryFor is a proxy of the class. If so, the solution here is just 

```js
return owner.factoryFor ? owner.factoryFor(type) : owner_lookupFactory(type);
```

Will report back once I have some answers.

jasonmit:
> @jasonmit what I meant is that we just force inject the owner on create no matter what the situation is since we have access to the owner regardless. That should simplify the problem for now until we get a better alternative.

Unsure, but I'd say that mimics the behavior of the new API versus embracing the new API which could lead to other unknowns.

>Im not at my computer right now so I will look into this more in a few hours. Im curious to see if the returned object from factoryFor is a proxy of the class. If so, the solution here is just

This will not work as the proxy throws by design: https://github.com/emberjs/rfcs/blob/master/text/0150-factory-for.md#detailed-design (See first bullet)

rwjblue:
FWIW, I am working on a polyfill now to provide the new API to older Ember versions.  Hope to have it published later today...

rwjblue:
https://github.com/rwjblue/ember-factory-for-polyfill should do the trick...

jasonmit:
PR updated with @rwjblue's polyfill.  If everyone is in agreement on it I'll write some tests.

addon/validations/factory.js
```diff
@@ -564,7 +563,7 @@ function getCPDependentKeysFor(attribute, model, validations) {
   let dependentKeys = validations.map((validation) => {
     let { options } = validation;
     let type = validation._type;
-    let Validator = type === 'function' ? BaseValidator : lookupValidator(owner, type);
+    let Validator = type === 'function' ? BaseValidator : lookupValidator(owner, type).class;
```
cibernox: Do you need class here? I think I don't fully grasp when the proxied object is enough and when it isn't.

rwjblue:
I'll keep working on this tonight, but I suspect that the polyfill doesn't support Ember < 2.3 yet (though I haven't tested).

addon/validations/factory.js
jasonmit: Yes, but I'm sending a PR into the polyfill to fix this.  Looks like discrepancy between polyfill and implementation within Ember.

When factory does not exist, undefined should be returned.  Currently, an empty FactoryManger object is returned.

/cc @rwjblue

jasonmit: https://github.com/rwjblue/ember-factory-for-polyfill/pull/6

jasonmit:
@rwjblue yup, it looks like <= 2.0 is failing

addon/validations/factory.js
jasonmit: Yes, the proxy does not proxy to the factory.  I linked above pointing to the RFC which explains that is by design.

https://github.com/emberjs/ember.js/blob/c05d241fee2b24de83d67da86bdbc5c8177d4266/packages/container/lib/container.js#L185

And the RFC (first bullet)
This will not work as the proxy throws (by design) when accessing anything other than class and create: https://github.com/emberjs/rfcs/blob/master/text/0150-factory-for.md#detailed-design (See first bullet)

cibernox: I understand. In this case you are not getting the factory to create an instance of that class, you want to read some information from the class itself.

addon/validators/base.js
```diff
@@ -89,11 +88,11 @@ const Base = Ember.Object.extend({
 
     if (!isNone(owner)) {
       // Since default error messages are stored in app/validators/messages, we have to look it up via the owner
-      errorMessages = factoryFor(owner, 'validator:messages');
+      errorMessages = owner.factoryFor('validator:messages');
     }
 
     // If for some reason, we can't find the messages object (i.e. unit tests), use default
-    errorMessages = errorMessages || Messages;
+    errorMessages = errorMessages && errorMessages.class ? errorMessages : Messages;
```
jasonmit: Once my [PR](https://github.com/rwjblue/ember-factory-for-polyfill/pull/6) into the polyfill is versioned, this line can be reverted.

rwjblue: Published as v1.0.2.

cibernox:
FWIW LGTM

offirgolan:
Build is failing for Ember <= 2.0. Looks like the polyfill isnt being injected at all. 

cc @rwjblue

rwjblue:
ZOMG, should be working now.  https://github.com/rwjblue/ember-getowner-polyfill/pull/18 + https://github.com/rwjblue/ember-factory-for-polyfill/compare/v1.0.2...v1.1.0.

We will need to bump to ember-getowner-polyfill@1.2.0 though.

offirgolan:
WOOT! Awesome job @rwjblue.

@jasonmit can you update the dependencies in this PR? Can we remove `ember-factory-for-polyfill` or do we need them both?

rwjblue:
Including only getowner polyfill _should_ work, but I'd definitely appreciate testing. The coupling is somewhat complicated, and it's definitely possible that I've mucked something up.

jasonmit:
@offirgolan PR ready.  Thanks.

offirgolan:
Released in `v3.2.3`. Thank you everyone for helping out on this 😸 !

Issue #193, opened 2016-05-23
Cleanup Docs


offirgolan:
I believe its time for a demo page / docs overhaul. A lot of issues have been opening up because of either lack of documentation or it just being confusing. This issue has been opened to track what needs to be changes / added.

- [ ] General cleanup (spelling / grammar mistakes)
- [ ] Add some better examples
- [ ] Maybe restructure to make things easier to find 
- [ ] New demo page with code snippets

blimmer:
Was thinking about the examples / demo page today, especially related to #200 . I think for the demo we should include a live-updating example (the validation is always there) and one where the validation appears on blur. Many people ask us about that and I think the answer to #200 is a good way to handle it and encourage use of one-way bindings / DDAU. what do you think @offirgolan ?

offirgolan:
@blimmer that sounds pretty solid. I was thinking of restructuring the demo like I have it in `ember-light-table` which has examples and expandable code snippets as well.

Issue #191, opened 2016-05-21
Unit tests fail after upgrading to 2.9.0 ("validator not found")


splattne:
After upgrading from 2.8.0 to 2.9.0 all my model unit tests fail with the following error message:

>Promise rejected after it exists: [ember-cp-validations] Validator not found of type: presence.@ 40 ms
>Source: 	
>Error: [ember-cp-validations] Validator not found of type: presence.
>     at lookupValidator (http://localhost:5200/assets/vendor.js:145633:13)
>        at http://localhost:5200/assets/vendor.js:145477:92
>        at Array.map (native)
>        at getCPDependentKeysFor (http://localhost:5200/assets/vendor.js:145474:37)
>        at createCPValidationFor (http://localhost:5200/assets/vendor.js:145355:25)
>        at http://localhost:5200/assets/vendor.js:145314:20
>        at Array.forEach (native)
>        at init (http://localhost:5200/assets/vendor.js:145313:31)
>        at superWrapper [as init] (http://localhost:5200/assets/vendor.js:33248:22)
>        at new Class (http://localhost:5200/assets/vendor.js:46980:14)

The tests are simple "it exists" tests which Ember CLI generates automatically, like 

    test('it exists', function(assert) {
      let model = this.subject();
      assert.ok(!!model);
    });

I reverted to 2.8.0 and the unit tests run successfully.

offirgolan:
Since the CPs now require container access, you will need to add all validators that your model depends on in your tests. 

```js
import { moduleFor, test } from 'ember-qunit';

moduleForModel('foo', 'Unit | Model | model', {
  needs: ['validator:presence']
});
```

Thanks for catching + reporting this! Ill add it to the release notes

offirgolan:
Im gonna go ahead and close this. Feel free to reopen if you face any similar issues 😺

PR #453, opened 2017-02-19
[DOCS] Add note on testing and container access to "Basic" documentation

BrianSipple:
This addresses part of #193 by adding a note on container access to the "Basic" section of the docs. 

Hopefully this helps any developers (myself included) who might be prone to encounter the issue noted in #191 😄.

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -92,8 +92,6 @@ import Validator from './validations/validator';
  * });
  * ```
  *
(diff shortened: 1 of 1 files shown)


offirgolan:
@BrianSipple this is great. Thank you. I've been meaning to add a note like that for some time now and never got a chance to.

PR #522, opened 2017-07-31
Update version of ember-getowner-polyfill

jrjohnson:
Since getOwner is always being pulled off of the Ember object updating
this to the latest version just removes the deprecation warning without
changing the behavior of the polyfill.


offirgolan:
Resolved in #523 😄

PR #532, opened 2017-08-29
remove bower usage

Dhaulagiri:
these deps are imported via npm

### DIFF
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,4 +1,0 @@
-{
-  "directory": "bower_components",
-  "analytics": false
-}
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,6 @@ dist: trusty
 
 cache:
   yarn: true
-  directories:
-    - $HOME/.cache # includes bowers cache
 
 addons:
   code_climate:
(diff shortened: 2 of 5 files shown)


offirgolan:
@Dhaulagiri bower is still used by ember-moment which is needed to run the test cases.

Dhaulagiri:
@offirgolan I felt like that wasn't true, particularly because ember-moment's blueprint [doesn't even add the deps to bower anymore](https://github.com/stefanpenner/ember-moment/pull/58).  In either case though, I don't think ember-cp-validations is actually using anything from ember-moment specifically is it? ember-cli-moment-shim, yes, but I don't see any uses of the template or computed helpers exposed by ember-moment.  If I remove ember-moment locally the test suite still passes.

jasonmit:
`ember-cli-moment-shim` no longer depends on bower for pulling in the moment lib(s).  If all `ember-cp-validations` is doing is importing `moment` (a quick grep shows that's probably the case) then you'll only need `ember-cli-moment-shim`.

Dhaulagiri:
I cleaned things up a little bit more based on @jasonmit's comment, leaving _only_ `ember-cli-shims` defined in this project's devDependencies.  Verified locally that the test suite passes (travis is borked for other reasons).

offirgolan:
Sorry guys I'm not following. @Dhaulagiri you removed ember-moment completely yet we can still access moment? Is there something I'm missing here?

Dhaulagiri:
I was confused about this as well this week, but let me see if I can clarify.

Currently this project has both `ember-moment` and `ember-cli-moment-shim` included.  The former provides some template and computed property helpers for working with moment, while the latter provides an ES6-accessible module for moment which lets you do `import moment from 'moment';` and work with moment directly. 

ember-cp-validations _only_ uses `moment` directly (from what I can tell), so there is no reason to include ember-moment since it is not actually being used.  While it _seems_ like `ember-moment` and `ember-cli-moment-shim` _have_ to be used together you can actually use the shim by itself if you want to.

offirgolan:
@Dhaulagiri thank you so much for the clarification. That makes a whole lot more sense 😄

PR #534, opened 2017-08-31
Attempt to fix the test suite

Dhaulagiri:
I'm not exactly sure what changed to make the test suite so unhappy, but in general ember-try seemed to be having problems installing ember data correctly so I updated a bunch of the failing scenarios to get them to work.

Of particular note:
 * It might make sense to remove some of the older scenarios at some point
 * I could not get the Ember 2.3 scenario working so I removed it.  Let me know if you feel strongly about keeping it and I can try again.
* The ember-release scenario was breaking because it couldn't run with ember data 2.14.  I think that is related to [this](https://github.com/emberjs/data/issues/5118) ember data issue so I made that scenario run with ember data 2.13 for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,13 +15,10 @@ addons:
 
 env:
   - EMBER_TRY_SCENARIO=default
-  - EMBER_TRY_SCENARIO=ember-1.11
-  - EMBER_TRY_SCENARIO=ember-1.12
   - EMBER_TRY_SCENARIO=ember-1.13
-  - EMBER_TRY_SCENARIO=ember-2.0
   - EMBER_TRY_SCENARIO=ember-lts-2.4
   - EMBER_TRY_SCENARIO=ember-lts-2.8
-  - EMBER_TRY_SCENARIO=ember-2.10
+  - EMBER_TRY_SCENARIO=ember-2.12
   - EMBER_TRY_SCENARIO=ember-release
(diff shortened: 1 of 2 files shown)


offirgolan:
@Dhaulagiri this is fantastic! Thank you so much for taking the time to fix this!

I 100% agree with you that some of these scenarios can be removed. What are your thoughts about the following:

```
env:
   - EMBER_TRY_SCENARIO=default
   - EMBER_TRY_SCENARIO=ember-1.13
   - EMBER_TRY_SCENARIO=ember-lts-2.4
   - EMBER_TRY_SCENARIO=ember-lts-2.8
   - EMBER_TRY_SCENARIO=ember-2.12
   - EMBER_TRY_SCENARIO=ember-release
   - EMBER_TRY_SCENARIO=ember-beta
   - EMBER_TRY_SCENARIO=ember-canary
```

Dhaulagiri:
That sounds good to me. It seems like 1.13 could be dropped as well at some point, but as long as 1.13 support isn't a hindrance it seem worth keeping around for now.

offirgolan:
Agreed. I plan to start dropping support for Ember < 2.0 when 3.0 is released.

Dhaulagiri:
@offirgolan updated the scenarios based on your suggestion

offirgolan:
@Dhaulagiri thank you! 😸

PR #537, opened 2017-09-08
fill warnings property on the validations object

### LINKED ISSUE #527: warnings are not filled on the validations object
### Environment

- Ember Version: 2.12.0
- Ember CLI Version: 
- Ember CP Validations Version: 3.4.0

### Steps to Reproduce

https://ember-twiddle.com/79f50d8ee340f865d535e809175997f4?openFiles=controllers.application.js%2C

@offirgolan  Do you have any idea why `warningMessages`, `isWarning` and `warning` are not set on the `{validations}` when calling `model.validate()`?

zidjian257:
Resolves #527  .

Changes proposed:

 - try to get warnings filled - change works, but don't know if it's the correct approach

### DIFF
--- a/addon/validations/result-collection.js
+++ b/addon/validations/result-collection.js
@@ -382,6 +382,11 @@ export default Ember.ArrayProxy.extend({
 
 function computeErrorCollection(attribute, content = []) {
   let errors = flatten(content.getEach('errors'));
+  let warnings;
+  if(errors.length === 0) {
+    warnings = flatten(content.getEach('warnings'));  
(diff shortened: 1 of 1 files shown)


offirgolan:
@zidjian257 sorry for the delay on this. I've been pretty swamped lately. 

I'll review this PR in the next couple days.

offirgolan:
@zidjian257 thanks for bringing this to my attention. The problem was a bit more complicated than I'd hoped for but I was able to come up with a much better solution in #538. If you can check out that branch and let me know if that fixed your issue, that would be great. 😸

PR #538, opened 2017-09-18
[BUGFIX] Rework Warning Logic

### LINKED ISSUE #527: warnings are not filled on the validations object

offirgolan:
Resolves #527.

The logic for the warning message implementation was incorrect. This PR fixes that by removing the `isWarning` flag and computing the warningMessage / message logic via the Result `update` method. 

Because there is now no flag to determine if a result is a warning or an error, I had to create a WarningResultCollection class which extends the ResultCollection class but reroutes all errors to warnings which solves the case of `belongs-to`, `has-many`, and `alias`. 

Tasks:

- [x] Added test case(s)
- [x] Updated documentation


zidjian257:
thanks for the fix - solves our issues mentioned in #537

PR #554, opened 2017-12-13
Adding Example of validator usage in component

swgamerx:
Adding a basic computed property to the object example that demonstrates how to get validation of non-model property via js.

Resolves # . 

I was looking at documentation in Usage/Basic and could not find an example of how to access a component property's validation. Found example in Validations/Accessing Validations. 

Changes proposed:

 - Adding Example of getting validation of a property in component to the existing documentation. 

Tasks:

- [x] Added test case(s) - _not applicable due to it just being a documentation update_
- [X] Updated documentation

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -88,7 +88,13 @@ import Validator from './validations/validator';
  * });
  *
  * export default Ember.Component.extend(Validations, {
- *   bar: null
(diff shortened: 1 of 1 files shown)


offirgolan:
Closing in favor of #564

PR #564, opened 2018-02-10
Update all the things

offirgolan:
Changes proposed:

 - Use ember-cli-update to update to the latest ember and ember-cli
 - Run codemod 
 - Setup eslint + prettier and run prettier
 - Support Ember 2.12 LTS+

Todo (in different PRs)

- ~~Fix ember beta env issue with the descriptor changes (@rwjblue would love some input on that)~~
- Get off of jquery 2.x (required by demo page which uses bootstrap3)


LICENSE.md
```diff
@@ -1,27 +1,9 @@
-Copyright 2016 Yahoo Inc.
```
rwjblue: Should probably bring back the original here

addon/-private/internal-result-object.js
```diff
@@ -1,29 +1,15 @@
-/**
```
rwjblue: are these not needed?

addon/utils/utils.js
```diff
@@ -33,7 +26,7 @@ export function unwrapProxy(o) {
 }
 
 export function isProxy(o) {
-  return !!(o && (o instanceof Ember.ObjectProxy || o instanceof Ember.ArrayProxy));
+  return !!(o && (o instanceof ObjectProxy || o instanceof ArrayProxy));
```
rwjblue: Might also want to check for `unknownProperty` (I know this is unrelated)

rwjblue:
> Fix ember beta env issue with the descriptor changes (@rwjblue would love some input on that)

Happy to chat through it, though I'm not sure off the cuff exactly what breaks for this ATM...

offirgolan:
> Happy to chat through it, though I'm not sure off the cuff exactly what breaks for this ATM

My concern with the new CP `get` rfc that landed in beta is when we try extract the dependentKeys from a CP passed to a validator. 

For example when you get a validator like this:

```js
{
  myProp: validator('presence, {
    presence: true,
    disabled: computed('model.foo', 'model.bar.@each.foo', function () {})
  });
}
```

We would extract the dependent keys from the disabled CP and add them to the `myProp` cp dependentKeys.

This currently happens [here](https://github.com/offirgolan/ember-cp-validations/blob/v3.5.1/addon/validations/factory.js#L617-L619).

It looks like the only thing that is breaking is the test case (since its trying to do someEmberObject.someProp.isDescriptor which throws since in the future someProp would return the computed value). 

Do you think the that logic we have now for extracting the dependentKeys is still okay?

LICENSE.md
offirgolan: Ended up reverting the license back.

PR #583, opened 2018-04-14
export hasValidations decorator for es6 style models

### LINKED ISSUE #556: Feature idea: decorator for adding validations to es6 style classes
Currently if you have your models written as es6 classes with decorators like this:

```js
import { attr, belongsTo } from 'ember-decorators/data';
import Validations from '../utils/validations/message';

export default class Message extends Model { // how to add Validations to class
  @belongsTo user
  @belongsTo bookingOrder
}
```

you'll be hard presseed to add the Validations mixin 

expect I made a decorator for this purpose that works like this

```js
import { attr, belongsTo, hasMany } from 'ember-decorators/data';
import Validations from '../utils/validations/message';
import { hasValidations } from "../utils/decorators";

@hasValidations(Validations)
export default class Message extends Model {
  @belongsTo user
  @belongsTo bookingOrder
```

I already wrote the validator (hasValidations) , so if anyone interested I can post it

danielspaniel:
Resolves #556 

Changes proposed:

 - export a new function ```hasValidations``` class level decorator which allows you to add validations to an es6 style ember data model or component 

Tasks:

- [x] Added test case(s)
- [x] Updated documentation

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,5 +1,6 @@
 module.exports = {
   root: true,
+  parser: 'babel-eslint',
   parserOptions: {
     ecmaVersion: 2017,
     sourceType: 'module'
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: node_js
 node_js:
   # we recommend testing addons with the same minimum supported node version as Ember CLI
   # so that your addon works for all apps
-  - "4"
+  - "6"
 
 sudo: false
 dist: trusty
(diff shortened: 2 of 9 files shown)


danielspaniel:
ember-beta has a bug that prevents build from succeeding when trying to test against it, so i put that in the tests that were allowed to fail.  This setting makes sense anyway .. because beta and canary are not usually stable, and to require them to pass for CI to pass seems a bit unreasonable.

offirgolan:
@danielspaniel this is awesome! Thanks for taking the time to get this done.

danielspaniel:
your welcome @offirgolan .. i am surprised no one did this before because you kinda need this if you make models in class based style

offirgolan:
They probably have but you're the only one to open a PR 😄

PR #586, opened 2018-04-19
update to 3.1 and fix ObjectProxy extends validations bug

danielspaniel:
Resolves an issue where you get this error:
 
Assertion Failed: You attempted to access the `__VALIDATIONS_MIXIN_COUNT__` property (of <(unknown mixin):ember391>).↵Since Ember 3.1, this is usually fine as you no longer need to use `.get()`↵to access computed properties. However, in this case, the object in question↵is a special kind of Ember object (a proxy). Therefore, it is still necessary↵to use `.get('__VALIDATIONS_MIXIN_COUNT__')`

when you try to extend ObjectProxy with validations and the content you are proxying is an ember-data model 

Changes proposed:

 - minor cleanup/update in ValidationsMixin within buildValidations to account for issues with extending ObjectProxy with Validations
- does not need documentation because it is merely fixing internal implementation
 
Tasks:

- [+ ] Added test case(s)


addon/validations/factory.js
```diff
@@ -101,8 +101,8 @@ export default function buildValidations(validations = {}, globalOptions = {}) {
       this._super(...arguments);
 
       // Count number of mixins to bypass super check if there is more than 1
-      this[VALIDATIONS_MIXIN_COUNT] = this[VALIDATIONS_MIXIN_COUNT] || 0;
-      validationMixinCount = ++this[VALIDATIONS_MIXIN_COUNT];
+      validationMixinCount = (this.get(VALIDATIONS_MIXIN_COUNT) || 0) + 1;
```
offirgolan: `get(this, VALIDATIONS_MIXIN_COUNT)`

addon/validations/factory.js
```diff
@@ -101,8 +101,8 @@ export default function buildValidations(validations = {}, globalOptions = {}) {
...
+      this.set(VALIDATIONS_MIXIN_COUNT, validationMixinCount);
```
offirgolan: `set(this, VALIDATIONS_MIXIN_COUNT, validationMixinCount)`

offirgolan:
@danielspaniel thank you for this PR 😸. Added 2 small comments but other than that, lgtm!

danielspaniel:
All set @offirgolan

offirgolan:
Thanks again @danielspaniel!

danielspaniel:
your welcome @offirgolan

Turbo87:
@offirgolan would be open to backporting this fix to the v3.x branch?

PR #587, opened 2018-04-24
Fix ObjectProxy extends validations bug (v3.x)

Turbo87:
Backport of #586 to the `v3.x` branch. I've opted to only backport the relevant change as the other things in #586 created a few complicated conflicts.

/cc @danielspaniel @offirgolan


offirgolan:
Released with `v3.5.3`

PR #598, opened 2018-07-21
Update has-many.js

JackEllis:
There was an issue with checking a hasMany for me, I'm not 100% sure on the exact circumstances but it was triggered after an upgrade to EmberJS 3.1+. Importing get, rather than assuming the object has the get method, fixed things nicely and is arguably a smart way to proceed anyway.

Edit: Hmm. I was trigger happy with this, it may not solve anything!


offirgolan:
Thanks @JackEllis!

offirgolan:
> Edit: Hmm. I was trigger happy with this, it may not solve anything!

Just saw this after I merged lol. Nonetheless, this PR is still totally fine. If you figure out what the issue is, just open open another PR 😄

PR #608, opened 2018-09-22
updated installation instructions for ember-intl support

st-h:
Resolves no issue. Was wondering when the outdated ember-cli-babel will finally be upgraded until I noticed that the package name seems to has been changed.

Changes proposed:

 - Updates the installation instructions to use @ember-intl/cp-validations instead of ember-intl-cp-validations

Tasks:

- [ ] Added test case(s)
- [x] Updated documentation


offirgolan:
Thanks @st-h 😄

PR #629, opened 2019-03-11
Enable parallel Babel builds

### LINKED ISSUE #628: `v-get` AST transform is preventing `ember-cli-babel` from parallelizing
`ember-cli-babel` and `ember-cli-htmlbars-inline-precompile` are able to parallelize their workload using `workerpool`. For that the plugin configurations need to be serializable and addons that register HTMLBars AST plugins need to slightly change the code to enable it.

One implementation example is https://github.com/rwjblue/ember-angle-bracket-invocation-polyfill/pull/45. Once we do the same thing for this addon here parallelizing should work fine.

```
DEBUG=ember-cli-htmlbars-inline-precompile* DEBUG_LEVEL=debug ember b
```
☝️ will enable debug output and show whether the transform was parallelizable or not

/cc @rwjblue

Turbo87:
Resolves #628

This will allow `ember-cli-htmlbars-inline-precompile` to parallelize itself, which will enable better build performance

/cc @offirgolan

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,14 +10,26 @@ module.exports = {
   name: 'ember-cp-validations',
 
   setupPreprocessorRegistry: function(type, registry) {
+    let plugin = this._buildPlugin();
+
+    plugin.parallelBabel = {
+      requireFile: __filename,
+      buildUsing: '_buildPlugin',
+      params: {},
+    };
+
+    registry.add('htmlbars-ast-plugin', plugin);
+  },
+
+  _buildPlugin() {
     var VGet = require('./htmlbars-plugins/v-get');
 
-    registry.add('htmlbars-ast-plugin', {
+    return {
       name: 'v-get',
       plugin: VGet,
       baseDir: function() {
         return __dirname;
       }
-    });
-  }
+    };
+  },
 };
--- a/yarn.lock
+++ b/yarn.lock
@@ -178,6 +178,13 @@ amd-name-resolver@0.0.7:
   dependencies:
     ensure-posix-path "^1.0.1"
 
+amd-name-resolver@1.2.0:
+  version "1.2.0"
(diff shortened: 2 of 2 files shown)


Turbo87:
CI seems to fail for Ember 1.13, but I have no idea how this change could affect just Ember 1.13. is that an unrelated failure?

offirgolan:
@Turbo87 thanks for the PR 😄. Why the changes in the lockfile? Is that due to a dedupe?

offirgolan:
> CI seems to fail for Ember 1.13, but I have no idea how this change could affect just Ember 1.13. is that an unrelated failure?

Looks like a minor hiccup. A rebuild fixed it.

offirgolan:
@Turbo87 is there a reason you've targeted the 3.x branch for this?

Turbo87:
@offirgolan we're still using the v3.x release series through `@ember-intl/cp-validations` and `ember-changeset-cp-validations`, so I was hoping we could release a new v3.x minor version with that feature since it's not a breaking change. otherwise, I'd have to wait for a stable v4.x release, update those other addons, and then finally be able to improve our build speed ;)

Turbo87:
> Why the changes in the lockfile?

The parallelization only works in more recent versions of `ember-cli-babel`. Those versions also have a `throwUnlessParallelizable` flag (see https://github.com/rwjblue/ember-angle-bracket-invocation-polyfill/pull/45/files) that I wanted to enable to make sure to not regress on this. Unfortunately, it seems that `ember-font-awesome` is not compatible with that yet, and that is used as a devDependency here. I figured the update was still worth it so I kept it in the PR.

offirgolan:
@Turbo87 I gotchu. `v3.5.5` released with this PR and I've cherrypicked your change into #630. I'll release a new beta with it as well 😄

Turbo87:
awesome, thanks a lot!

PR #633, opened 2019-04-12
Do not override _result

mydea:
Overwriting `_result` triggers a deprecation warning in newer versions of Ember.
This PR fixes this by instead writing another overwrite-property, which is used by _result instead (if set).


mydea:
The failing tests are unrelated to this change, I believe.

PR #636, opened 2019-04-18
[WIP] Alon/fix decorators

### LINKED ISSUE #635: Error when using computed values with ember-decorators-polyfill installed or Ember 3.10.0
### Environment

- Ember Version: 3.6 with ember-decorators-polyfill or current 3.10.0-beta.2
- Ember CLI Version: 3.6
- Ember CP Validations Version: 3.5.5 and 4.0.0-beta.7 

### Steps to Reproduce
I tested this with my work app at 3.8 and used the dummy app at 3.6 and installed ember-decorators-polyfill

Having validations that use a form of `computed` for their value.
models/user-detail.js - dob validator
```
after: computed(function() {
  return moment()
    .subtract(120, 'years')
    .format('M/D/YYYY');
}).volatile(),
```
Results in the error:
Error: Assertion Failed: EmberObject.create no longer supports defining computed properties. Define computed properties using extend() or reopen() before calling create().

I tracked it in the debugger to 
-private/options.js in the constructor for the Options class
```
constructor({ model, attribute, options = {} }) {
  const optionKeys = keys(options);
  const createParams = { [OPTION_KEYS]: optionKeys, model, attribute };

  // If any of the options is a CP, we need to create a custom class for it
  if (optionKeys.some(key => isDescriptor(options[key]))) {
    return OptionsObject.extend(options).create(createParams);
  }

  return OptionsObject.create(createParams, options);
}
```

I chatted with Chris Garrett in e-decorators on the Ember discord. He suggested that the computed properties were not being detected and that you would always need to extend. He said the same error would apply with Ember 3.10 and I have been able to reproduce it with the dummy app and ember-source upgraded to 3.10.0-beta.2 

I attempted update the code to always return `OptionsObject.extend(options).create(createParams);` 

Which got rid of the error but then some of the tests were not passing.

Please let me know if there is more detail that I can provide.

Alonski:
Resolves #635 .

Changes proposed:
 - Update Ember to v3.8
 - Switched decorator usage to `ember-decorators-polyfill`
 - Fixed `isDescriptor` in utils to handle new decorator syntax
 - Fixed Native Classes and Decorator Usage
 - Updated ESLint and Prettier plugins to support decorator syntax
 - Fixed new Prettier lint rules regarding EOL

Currently Tests are failing :)

Tasks:

- [ ] Added test case(s)
- [ ] Updated documentation


README.md
```diff
@@ -44,7 +44,8 @@ If you are upgrading from 3.x to 4.x, please checkout the [upgrading documentati
```
offirgolan: Please remove all changes to this file as they are unnecessary.

tests/dummy/app/models/company.js
```diff
@@ -1,16 +1,18 @@
...
   buildValidations
 } from 'ember-cp-validations';
 
+const { attr, Model } = DS;
+
 export const Validations = buildValidations({
   name: validator('presence', { presence: true, description: 'Name' })
 });
 
+export default
```
offirgolan: Should be 

```js
@hasValidations(Validations)
export default class Company extends Model {}
```

I believe this is why the tests are failing.

package.json
```diff
@@ -101,6 +98,7 @@
   },
   "dependencies": {
     "ember-cli-babel": "^7.1.2",
+    "ember-decorators-polyfill": "^1.0.3",
```
offirgolan: This addon doesnt need a hard dependency on `ember-decorators-polyfill` since we just export a decorator function with no dependencies.

This can be moved into devDeps.

GavinJoyce:
@Alonski thanks for working on this, I'm hitting this issue while upgrading my app to `3.10` too. Do you need any help with it?

Pixelik:
Hello 👋 

Our library is currently broken because we test it against future ember releases and we get `Error: Assertion Failed: EmberObject.create no longer supports defining computed properties` when we do so. I believe this PR is supposed to fix it and I was wondering whether this PR is still ongoing or close to being approved? 😁 

Thanks 🙏

Alonski:
@GavinJoyce @Pixelik I'll try to fix this up tomorrow.
@offirgolan Thanks for the review!

tests/dummy/app/models/company.js
Alonski: Fixed this but tests are still failing

Alonski:
@GavinJoyce I would love some help here. I have managed to "fix" a few tests by declaring the `dependantKeys` as an array on the buildValidations object like this:
```js
    let Validations = buildValidations(
      {
        firstName: {
          description: readOnly('model.description'),
          validators: [
            validator('length', {
              min: alias('model.minLength')
            })
          ]
        }
      },
      {
        disabled: not('model.enabled')
      },
      {
        dependentKeys: ['model.enabled', 'model.minLength', 'model.description']
      }
    );
```
This is less than ideal for the long run but at least it got a few tests passing.

The bigger issue here, and why the last 16 tests are failing, is in the way that mixins and inheritance work.
It seems that multiple mixins or inheritence don't merge validations correctly.

I think the issue could be here: https://github.com/Alonski/ember-cp-validations/blob/5211784d2b695d0c541f8473daaf80059a4754a8/addon/validations/factory.js#L656

`option._dependentKeys` is always undefined.
I am not sure how to proceed here. It seems that the whole architecture for this addon needs to changed going forward with the direction Ember is heading. 

Again I would love some help from anyone will to help figure this out :)
Thanks @offirgolan

jderr-mx:
One thought I had about the options dependent keys issue would be to create an assertion to tell the user that the model's dependent keys need to be added to the dependentKeys array of the Validation.

It is not ideal, but would notify the user that need to use this workaround if they are using an affected version or the polyfill.

GavinJoyce:
> option._dependentKeys is always undefined.

I [searched emberobserver.com for `_dependentKeys`](https://www.emberobserver.com/code-search?codeQuery=_dependentKeys) which turned up this PR which moves away from using `_dependentKeys`:

https://github.com/kellyselden/ember-macro-helpers/pull/258

I'm not yet sure what our way forward is yet though

GavinJoyce:
Some more useful info here: https://github.com/emberjs/ember.js/issues/17641

GavinJoyce:
Screen Shot 2019-05-20 at 15 37 11

GavinJoyce:
Reading the [docs](http://offirgolan.github.io/ember-cp-validations/docs/modules/Advanced.html#computed-options), it seems that validators accept option hashes which can contain CPs:

```js
{
  disabled: computed.readOnly('model.disableValidations')
}
```

`All options can also be Computed Properties. These CPs have access to the model and attribute that is associated with the validator.`

**A question for someone who's familiar with the internals of this addon, prompted by @rwjblue's comment above**: can we use the `disabled` computed property as a dependency instead directly instead of using its dependent keys?

Alonski:
@offirgolan Maybe you can shine some light on ☝️

GavinJoyce:
In the interests of being pragmatic and making progress, I'm going to create a spike which uses private ember internals to get the computed property keys. I believe that this should be possible

GavinJoyce:
I've managed to get a green build on Ember 3.8 and 3.10 for the `dependantKey` issue:

https://github.com/GavinJoyce/ember-cp-validations/pull/1

I'm going to try this in my own app's Ember 3.10 WIP branch now

simonihmig:
@GavinJoyce I tried your branch in our app, and indeed it seems to fix the issue (using 3.11-canary)!

Alonski:
Spoke to Gavin and if we want to merge in his fix we will probably need to do some Ember version checking to see if we can use it or if we need to revert to the old ways...

If no one picks this up in a few days I will try to get to it.

alejandrodevs:
@GavinJoyce I tried your branch in our app and everything is working well.

offirgolan:
Closing because of #641

Issue #640, opened 2019-05-15
Broken with Ember 3.10.0


stopfstedt:
### Environment

- Ember Version: 3.10.0
- Ember CLI Version: 3.10.0
- Ember CP Validations Version: 4.0.0-beta.8

### Steps to Reproduce

```
npx ember-cli-update --to=3.10.0
npm install
npm test
```

Fails with error message: `EmberObject.create no longer supports defining computed properties. Define computed properties using extend() or reopen() `

refs #635

patocallaghan:
FYI this issue has started happening again with the Ember 3.13 beta but I have a fix here https://github.com/offirgolan/ember-cp-validations/pull/653

PR #653, opened 2019-08-21
Use correct path to access ember metal package in Ember 3.13

patocallaghan:
Related to #640 

It appears the assertion `EmberObject.create no longer supports defining computed properties. Define computed properties using extend() or reopen()` has re-appeared in Ember 3.13 beta.

To fix the problem in 3.10, logic was added to grab the Ember Metal module and use that to determine when a property is a computed property.

In Ember 3.13-beta it looks like the path for Ember Metal is no longer `@ember/-internals/metal` but `@ember/-internals/metal/index` instead.

### 3.12
image

### 3.13 beta
image

### DIFF
--- a/addon/-private/ember-internals.js
+++ b/addon/-private/ember-internals.js
@@ -1,8 +1,13 @@
 import Ember from 'ember';
 
 let __EMBER_METAL__;
-if (Ember.__loader.registry['@ember/-internals/metal']) {
-  __EMBER_METAL__ = Ember.__loader.require('@ember/-internals/metal');
+let emberMetalPaths = [
+  '@ember/-internals/metal', // ember-source from 3.10
+  '@ember/-internals/metal/index' // ember-source from 3.13
+];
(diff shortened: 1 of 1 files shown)


patocallaghan:
Looks like CI is failing with a node engine related error 

image

@offirgolan would you open to a PR which bumped upgraded ember-cli to 3.12. Seeing as the major breaking change release is still in beta this would be a good opportunity to also drop support for Node 6 (which is no longer in ember-cli 3.12)

patocallaghan:
I've created a PR which upgrades Ember CLI to 3.12 and drops support for Node 6 https://github.com/offirgolan/ember-cp-validations/pull/654

jrjohnson:
@offirgolan [Ember 3.13 came out yesterday](https://github.com/emberjs/ember.js/releases/tag/v3.13.0) and this is going to start breaking in many places. Any chance it can be merged and released?

offirgolan:
Closing in favor of #668.

Issue #642, opened 2019-05-29
Failing tests with ember 3.11 beta


GavinJoyce:
The `beta` and `canary` scenarios are failing in https://github.com/offirgolan/ember-cp-validations/pull/641 with the following errors:

```
not ok 46 Chrome 74.0 - [38 ms] - Integration | Validations | Factory - General: call super in validations class with no super property
    ---
        actual: >
            null
        stack: >
            Error: Assertion Failed: An EmberObject based class, (unknown), was not instantiated correctly. You may have either used `new` instead of `.create()`, or not passed arguments to your call to super in the constructor: `super(...arguments)`. If you are trying to use `new`, consider using native classes without extending from EmberObject.
                at assert (http://localhost:7357/assets/vendor.js:43163:17)
                at Class.CoreObject (http://localhost:7357/assets/vendor.js:36333:32)
                at Class.EmberObject (http://localhost:7357/assets/vendor.js:37242:26)
                at Class.DebugFrameworkObject (http://localhost:7357/assets/vendor.js:37334:32)
                at Class (http://localhost:7357/assets/vendor.js:36699:24)
                at new Class (http://localhost:7357/assets/vendor.js:36699:24)
                at Function.create (http://localhost:7357/assets/vendor.js:36766:20)
                at _default (http://localhost:7357/assets/tests.js:242:16)
                at Object.<anonymous> (http://localhost:7357/assets/tests.js:1504:49)
                at runTest (http://localhost:7357/assets/test-support.js:4400:30)
```

fsmanuel:
@GavinJoyce Is this still a problem with the latest beta release?

PR #654, opened 2019-08-21
Update addon to Ember CLI 3.12

patocallaghan:
This is a follow on to my PR at https://github.com/offirgolan/ember-cp-validations/pull/653. It appears [CI is failing](https://travis-ci.org/offirgolan/ember-cp-validations/jobs/574978861) due to a node 6 engine incompatibility issue. 

Seeing as we're currently in a 4.x beta period it's probably worth upgrading the addon to ember-cli 3.12 which also drops Node 6 support.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,7 +2,7 @@ module.exports = {
   root: true,
   parser: 'babel-eslint',
   parserOptions: {
-    ecmaVersion: 2017,
+    ecmaVersion: 2018,
     sourceType: 'module'
   },
   plugins: ['ember'],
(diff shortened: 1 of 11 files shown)


patocallaghan:
A single test fails in CI: `Integration | Validations | Factory - General: call super in validations class with no super property`

The error being:

> Error: Assertion Failed: An EmberObject based class, (unknown), was not instantiated correctly. You may have either used `new` instead of `.create()`, or not passed arguments to your call to super in the constructor: `super(...arguments)`. If you are trying to use `new`, consider using native classes without extending from EmberObject.

This failing test is unrelated to this PR as this has been failing since 3.11. There's an already an issue for it in https://github.com/offirgolan/ember-cp-validations/issues/642

offirgolan:
Closing in favor of #668.

Issue #659, opened 2019-09-17
[ember-source 3.13] Infinite loop causes stack overflow


patocallaghan:
### Environment

- Ember Version: 3.13.0-beta.5
- Ember CLI Version: 3.6.1
- Ember CP Validations Version: `master`
- Reproduction PR: https://github.com/offirgolan/ember-cp-validations/pull/660

While preparing our app to be Ember 3.13-ready we've come across a number of infinite loops which cause our app to crash. I'm not exactly sure what the problem is but I think it has to do with the recent `@tracked` internal changes which are shipping with 3.13. The failing tests I've created in https://github.com/offirgolan/ember-cp-validations/pull/660 here are passing in 3.12.0 but start failing when I bump to 3.13.0-beta.1.

### Steps to Reproduce
I have a component called `<MyComponent />`. It has a computed property which is based on `cheese` which can be passed in and some validations on a passed in model, `model.validations.isValid`. The validations are nothing special and in my reproduction case are just using what where being used for tests in the ember-cp-validations repo.

```js
import Component from '@ember/component';
import { and } from '@ember/object/computed';

export default Component.extend({
  myProp: and('cheese', 'model.validations.isValid')
});
```

```hbs
{{this.myProp}}
```

If `cheese` is `undefined` it seems to cause an infinite loop in the `model.validations.isValid` part of the computed property.

```
Uncaught (in promise) RangeError: Maximum call stack size exceeded
    at setupMandatorySetter (index.js:753)
    at tagForProperty (index.js:239)
    at ComputedProperty.get (index.js:3302)
    at Class.CPGETTER_FUNCTION [as username] (index.js:2639)
    at getPossibleMandatoryProxyValue (index.js:1764)
    at get (index.js:1837)
    at _getPath (index.js:1871)
    at Object.get (index.js:1821)
    at factory.js:489
    at Array.map (<anonymous>)
```

`console.log`-ing what is output on the `_getPath` line shows that it seems to be stuck in some sort of loop trying to retrieve the validation for the `username`, before it throws a stack overflow

![image](https://user-images.githubusercontent.com/685034/65058456-2dd2dd00-d96c-11e9-9599-73ddf7c9beeb.png)

Strangely, doing a few things here cause the infinite loop not to occur and the test to pass fine.

[...]

PR #660, opened 2019-09-17
Reproduction of 3.13.0 infinite loop in #659

patocallaghan:
This reproduces the infinite loop issue I opened in https://github.com/offirgolan/ember-cp-validations/issues/659

I've added two types of component:

1. `my-component` - js + template, Infinite loop is triggered from the template
2. `another-component` - js only. Infinite loop is triggered from the `init` hook

`yarn ember test --server --filter="Integration | Component | my-component: it renders"`

`yarn ember test --server --filter="Unit | Component | another-component: it creates correctly"`

Also worth noting while there is PR to make ember-cp-validations compatible with 3.13 in https://github.com/offirgolan/ember-cp-validations/pull/653 I find this issue is still happening in spite of cherry-picking that fix. So I opted to keep this PR smaller by only bumping to ember-source 3.13.0-beta.5

Edit: I've also added a reproduction of this happening in a running app (see https://github.com/offirgolan/ember-cp-validations/issues/659#issuecomment-532552985). You can see the effects by running `yarn start` and visiting http:localhost:4200.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "ember-qunit": "^3.4.1",
     "ember-qunit-nice-errors": "^1.2.0",
     "ember-resolver": "^5.0.1",
-    "ember-source": "~3.8.0",
+    "ember-source": "~3.13.0-beta.5",
     "ember-source-channel-url": "^1.1.0",
     "ember-test-selectors": "^0.3.8",
     "ember-truth-helpers": "2.0.0",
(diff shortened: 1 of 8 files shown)


patocallaghan:
Closing as a fix has being released in https://github.com/emberjs/ember.js/releases/tag/v3.13.0

PR #662, opened 2019-09-23
[WIP] Support Ember 3.13.0

GavinJoyce:
This PR adds support for Ember 3.13.0

Screenshot 2019-09-23 at 13 27 18

There is one issue which might need to be resolved before this is merged:

 * [ ] https://github.com/offirgolan/ember-cp-validations/issues/642 - I'm currently skipping this test. @snewcomer, perhaps you might have context on the need for this test


addon/-private/ember-internals.js
```diff
@@ -1,16 +1,21 @@
 import Ember from 'ember';
 
 let __EMBER_METAL__;
-if (Ember.__loader.registry['@ember/-internals/metal']) {
-  __EMBER_METAL__ = Ember.__loader.require('@ember/-internals/metal');
+let emberMetalPaths = [
```
efx: if this lands we should close #653.

offirgolan:
>  #642 - I'm currently skipping this test.

The test is related to this part of the code base: 

https://github.com/offirgolan/ember-cp-validations/blob/2e901724390172d3b9584aec69696e2f8e28fc29/addon/validations/factory.js#L112-L117

And is pretty much testing that if there is no super class with a validations mixin, it shouldnt call `this._super` at all. This was related to a bug that was found some time ago that @rwjblue helped solve. Pretty much, if you instantiated the validations object within an ember action, it will call `this._super` on that action which would invoke that action's parent method.

offirgolan:
We can skip the test for now and come back for this later if it means we can merge this and unblock everyone else.

offirgolan:
@GavinJoyce is this PR still WIP?

.travis.yml
```diff
@@ -3,7 +3,7 @@ language: node_js
 node_js:
   # we recommend testing addons with the same minimum supported node version as Ember CLI
   # so that your addon works for all apps
-  - '6'
+  - '10'
```
rwjblue: We should probably use 8 (minimum supported Node) and do this in a separate PR to drop Node 6 support (will be breaking). Will also need to tweak the `engines.node` entry in `package.json`.

addon/-private/ember-internals.js
```diff
@@ -1,16 +1,21 @@
...
+  '@ember/-internals/metal', // ember-source from 3.10
+  '@ember/-internals/metal/index' // ember-source from 3.13
+];
+let metalPath = emberMetalPaths.find(path => Ember.__loader.registry[path]);
```
rwjblue: FWIW, we shouldn't use `Array.prototype.find` if we still intend for this library to support IE11 (I'm not sure if we do though?).

Possible future refactor would be to swap this to using `ember-compatibility-helpers` so we don't need iterative looping.

addon/-private/ember-internals.js
```diff
@@ -1,16 +1,21 @@
...
+  __EMBER_METAL__ = Ember.__loader.require(metalPath);
 }
 
 export function getDependentKeys(descriptorOrDecorator) {
   if (__EMBER_METAL__ && __EMBER_METAL__.descriptorForDecorator) {
     let descriptor = __EMBER_METAL__.descriptorForDecorator(
       descriptorOrDecorator
     );
-    return descriptor._dependentKeys;
+    return descriptor._dependentKeys || [descriptor.altKey];
```
rwjblue: What scenario is `altKey` used for?

GavinJoyce: I'm not sure TBH. There are failing tests without it

addon/validations/factory.js
```diff
@@ -633,6 +633,8 @@ function getCPDependentKeysFor(attribute, model, validations) {
     dependentKeys.push('model.isDeleted');
   }
 
+  dependentKeys = dependentKeys.filter(dependentKey => !!dependentKey);
```
balinterdi: `dependentKeys.filter(Boolean)` ?

.travis.yml
Turbo87: to get this unblocked in a non-breaking way we could also run the scenarios for Ember 3.13 and above on `node_js: 10` and keep the others at `node_js: 6` for now until we actually drop support

jrjohnson:
This is our blocker for trying out the Octane preview. Anything I can do to help move this along?

.travis.yml
jrjohnson: I went with @Turbo87's suggestion and pinned the new scenarios at `8`. I'll follow up with another to drop `Node 6`.

addon/-private/ember-internals.js
jrjohnson: Papered this over with `Ember.A` for now.

jrjohnson:
Rebased in #668

PR #664, opened 2019-09-26
Fix CI builds

Turbo87:
Ember 3.13 and above are no longer supporting Node.js 6, but since this project still supports it and dropping support would be a breaking change we need to adjust our CI pipeline to use Node.js 8 instead for the ember-try scenarios that need it.

(as suggested in https://github.com/offirgolan/ember-cp-validations/pull/662#discussion_r328584709)

/cc @GavinJoyce @rwjblue @offirgolan


Turbo87:
It looks like the only things that are now legitimately failing CI are the changes in Ember 3.13 like:

> Assertion Failed: EmberObject.create no longer supports defining computed properties. Define computed properties using extend() or reopen() before calling create().

Those are addressed by #662 AFAIK

Turbo87:
closed in favor of #667

PR #667, opened 2019-10-14
[Breaking] Drop Node 6

jrjohnson:
v6 was end of life in April 2019 and is no longer supported.

This is required for #662


jrjohnson:
@offirgolan I propose that this be merged even with the failing tests as those scanrios are legit failures that will be fixed in #662.

jrjohnson:
@offirgolan thoughts on this? Would you rather I added it to #662?

offirgolan:
Lgtm, thanks for opening this PR 😄

PR #668, opened 2019-10-20
Support Ember 3.13 (redux)

jrjohnson:
Re-bases and Supersedes and closes #662 and #653

I couldn't figure out a practical way to keep helping move this along in the original branch since it needed a rebase so I created a new one.  Hope that is ok @offirgolan. I believe this is the last thing needed to support Ember 3.13 moving forward.


addon/validations/factory.js
```diff
@@ -633,6 +633,8 @@ function getCPDependentKeysFor(attribute, model, validations) {
     dependentKeys.push('model.isDeleted');
   }
 
+  dependentKeys = dependentKeys.filter(dependentKey => !!dependentKey);
```
offirgolan: This can be combined with the map below.

```js
dependentKeys = dependentKeys.filter(Boolean).map(...)
```

jrjohnson: Do you mean as: 
```
dependentKeys = dependentKeys
    .filter(dependentKey => !!dependentKey)
    .map(d => {
      return d.replace(/^model\./, `${ATTRS_MODEL}.`);
    });
```

offirgolan:
```js
dependentKeys = dependentKeys
    .filter(Boolean)
    .map(d => d.replace(/^model\./, `${ATTRS_MODEL}.`));
```

jrjohnson: Ah. Got it. Done!

jrjohnson:
@offirgolan tests are passing 🎉

tests/integration/validations/factory-general-test.js
```diff
@@ -1040,7 +1040,8 @@ module('Integration | Validations | Factory - General', function(hooks) {
     assert.equal(child.get('validations.errors.length'), 1);
   });
 
-  test('call super in validations class with no super property', function(assert) {
+  // https://github.com/offirgolan/ember-cp-validations/pull/656
+  skip('call super in validations class with no super property', function(assert) {
```
Turbo87: do we not want to address this before moving forward here?

jrjohnson: Decision from @offirgolan was to [skip for now](https://github.com/offirgolan/ember-cp-validations/pull/662#issuecomment-534187453). At this point getting this updated for Ember 3.13 is getting past critical as it's a blocker for many apps to provide feedback on the Octane beta.

Turbo87: okay... I would just be worried about apps not working correctly if we skip this test. 🤔

jrjohnson: Me too, I don't actually know what any of this stuff does! I'm just playing referee to get the approved path merged. I'm happy to investigate and do more, but getting this merged has been a *very* long process so far. Makes me very hesitant to add any more complexity that could delay it further.

offirgolan: @Turbo87 skipping as this is already broken on the latest beta. Will have to circle back and figure out a good solution for this.

offirgolan:
Released in `v4.0.0-beta.10`. Thank you for seeing this through @jrjohnson.

jrjohnson:
Yay! Thanks!!

PR #712, opened 2022-06-24
Lint everything

fsmanuel:
This is a followup on #711 to get CI green.

Most of the commit messages should be self explanatory and it should be easy to review commit by commit.
Failing scenarios: The release, beta and canary runs still fail. The error is:
```
broccoliBuilderErrorStack: TypeError: Cannot read property 'visitor' of undefined
```
and is at least caused by the `htmlbars-plugins/v-get.js` template transform and maybe other dependencies.

I would suggest that I open a followup PR with the changes to the template transform (which I already did - 7f986309cde3c5a8a9ecb07471b3553923dd4803) to keep this PR focused.


fsmanuel:
@NullVoxPopuli What do you think which versions of ember should we support? As you already mentioned here: https://github.com/adopted-ember-addons/ember-cp-validations/pull/711#discussion_r906057181 the new release should be a major bump.
The versions from `2.16` - `3.12` fail because of dependencies so I would suggest we drop all versions `< 3.16` and I'll see that I get `release`, `beta` and `canary` working by adding `ember-auto-import` `>= 2.0` to the try config.

NullVoxPopuli:
yup, sounds good to me. I don't think it makes sense to support ember pre-octane

fsmanuel:
@NullVoxPopuli I updated the PR description ^^ and would suggest that I open a followup PR with the changes to the template transform (which I already did) to keep this PR focused.

fsmanuel:
@NullVoxPopuli Thanks for merging. Wanna add the internal label?

PR #718, opened 2022-08-17
Deprecate volatile option

fsmanuel:
Ember 4.0 removed `.volatile()` on `computed()`. This PR adds the deprecation to remove it in version `5.0` of this addon.

Tests for ember-4.0, release, beta, and canary are expected to fail.

### DIFF
--- a/addon/validations/factory.js
+++ b/addon/validations/factory.js
@@ -7,6 +7,7 @@ import { cancel, debounce as debounced } from '@ember/runloop';
 import { guidFor } from '@ember/object/internals';
 import { isEmpty, isNone } from '@ember/utils';
 import { getOwner } from '@ember/application';
+import { deprecate } from '@ember/debug';
 import deepSet from '../utils/deep-set';
 import ValidationResult from '../-private/result';
 import ResultCollection from './result-collection';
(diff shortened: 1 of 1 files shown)


fsmanuel:
@MelSumner before we release a new beta version we should merge this to get the deprecation out.

rwwagner90:
@fsmanuel we should just remove this entirely. I wouldn't worry about deprecating it. We should update all the things to work with latest Ember and not worry about supporting legacy Ember versions IMO.

rwwagner90:
@fsmanuel yes, but `volatile` has been deprecated a very long time in Ember. I think we can remove it. Those who want to still use it can just use earlier versions of this addon.

knownasilya:
Agreed. Could you remove it @fsmanuel? I can merge and release once that's done.

MelSumner:
@knownasilya I agree with removing it instead of adding a deprecation notice, but should we get it working for 4.0 before doing a new release? WDYT?

knownasilya:
I meant the next version for the beta. I think there was a few things still left to remove the beta flag.

fsmanuel:
@rwwagner90 @MelSumner @knownasilya My point is that `volatile` is a documented option/feature of the addon and if we remove it things will break without a notice. So I thought we deprecate it and add an `assert` when we remove it.

If you disagree I'm fine with it and we can close the PR and release a new beta version.

I'll continue working on #717 and drop `volatile` there.

knownasilya:
@fsmanuel yeah. Let's do that. We can do multiple major versions if we need.

addon/validations/factory.js
```diff
@@ -407,6 +408,19 @@ function createAttrsClass(validatableAttributes, validationRules, model) {
  */
 function createCPValidationFor(attribute, model, validations) {
   let isVolatile = hasOption(validations, 'volatile', true);
+
+  deprecate(
+    '[ember-cp-validations] The `volatile` option should no longer be used ' +
+      'as it was removed in ember 4.0',
```
knownasilya, line 414: Should probably suggest an alternative. Maybe something like: Instead you can use a getter to accomplish the same thing.

fsmanuel: The problem is that all validators are computed properties. To accomplish the same thing you would need to have a dependent key that is for example a timestamp that always changes...

knownasilya: Yeah, you are right. Been a while since I used them.

PR #731, opened 2023-05-30
Update ember-cli to v4.4 (drops node v12)

fsmanuel:
- Updates ember-cli to v4.4
- Updates ember-data to v4.4
- Updates and drops dependencies
- Replaces `ember-font-awesome` with `@fortawesome/ember-fontawesome`

There will be a follow up PR with the update to ember-cli v4.8 as that is more involved and needs changes for ember-data to work.

Release, Beta, and Canary are expected to fail. Current test coverage is `3.28` -> `4.4`

https://github.com/adopted-ember-addons/ember-cp-validations/issues/716


knownasilya:
Could you add ember 4.4 to ember-try/actions configs?

package.json
```diff
@@ -50,60 +50,60 @@
...
-    "release-it-lerna-changelog": "^3.1.0",
-    "sass": "^1.15.3",
-    "webpack": "^5.65.0",
-    "yuidoc-ember-theme": "^2.0.1"
+    "release-it": "^15.10.3",
+    "sass": "^1.62.1",
+    "webpack": "^5.72.1"
   },
   "engines": {
     "node": "12.* || 14.* || >= 16"
```
knownasilya, line 109: Was node 12 dropped? Noticed you bumped above

fsmanuel:
@knownasilya We don't need to add 4.4 as it's the default (in package.json) and tested via the normal tests. It seems `ember-cli-update` takes good care of that and the next update to ember-cli v4.8 will add the `ember-lts-4.4` scenario.

package.json
fsmanuel: No, that will happen in ember-cli `v4.8`. I think it was a `devDependency` that needed `v12` already.

fsmanuel: If you put the breaking label because of node v12 I don't think we need it. But the next PR will be breaking. I'll highlight that in the PR.

.github/workflows/ci.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - name: Install Node
         uses: actions/setup-node@v3
         with:
-          node-version: 12.x
```
knownasilya, line 25: I'd revert these if node 12 isn't being removed

fsmanuel: So the problem is that `@ember/test-helpers@^2.7.0` has a dependency on `@embroider/util@^1.9.0` that resolves to `@embroider/util@1.11.0`.

```sh
error @embroider/util@1.11.0: The engine "node" is incompatible with this module. Expected version "14.* || >= 16". Got "12.22.12"
```

What should we do now?

knownasilya: I think it's fine to drop node 12, support has ended for it a while back (on the node front).

fsmanuel: Alright. Droped it and added the breaking label again.

knownasilya (approved):

PR #738, opened 2023-10-26
Add ember-data v4.12 to test coverage

fsmanuel:
We don’t change the package node requirements so it’s not a breaking change but to test against ember-data v4.12 we need node v16 in CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Install Node
         uses: actions/setup-node@v3
         with:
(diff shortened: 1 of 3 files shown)


mansona:
if we're not testing against Node 14 then it's going to be impossible to maintain support for it. I ended up bumping the default ember-data version in https://github.com/adopted-ember-addons/ember-cp-validations/pull/743 but got around the engines issue by using `--ignore-engines` in the workflow 👍

fsmanuel:
@mansona nice!

PR #744, opened 2023-11-08
Remove Ember. from doc strings

mansona:
This is a cherry-pick of work done by @fsmanuel in https://github.com/adopted-ember-addons/ember-cp-validations/pull/736

I'm just extracting the changes because they are individually important and not related to updating any more 👍


fsmanuel (approved):
Thanks! That makes a lot of sense.

PR #748, opened 2023-11-08
fix peer-dependencies

mansona:
I'm not exactly sure what happened 🤔 but in merging some of the recent PRs we ended up with 2 `peerDependencies` blocks 🤷 this PR fixes that 👍


fsmanuel (approved):
Good catch!

PR #749, opened 2023-11-08
update @embroider/test-setup to fix CI

mansona:
The `@embroider/test-setup` package was very out of date so we really couldn't expect the `embroider-safe` or `embroider-optimised` tests to pass without updating it 👍


mansona:
Using @Pixelik's suggestion in https://github.com/adopted-ember-addons/ember-cp-validations/issues/733#issuecomment-1719467904 brings the failing tests down from 9 to 5 which is progress I guess 🤷

PR #755, opened 2024-02-26
add no-deprecations ember-try scenario

mansona:
This is using https://github.com/mansona/ember-deprecation-error to try to get a better error message to see why the `ember-release` job is actually failing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,6 +72,7 @@ jobs:
           - ember-canary
           - embroider-safe
           - embroider-optimized
+          - no-deprecations
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


mansona:
So it looks like this is failing because of an ember-data deprecation: 

```The filter method on ember-data's PromiseManyArray is deprecated. await the promise and work with the ManyArray directly.```


In the other PR @runspired mentioned that we should probably just throw [here](https://github.com/adopted-ember-addons/ember-cp-validations/blob/main/addon/utils/utils.js#L77) if we have a PromiseManyArray https://github.com/adopted-ember-addons/ember-cp-validations/pull/735#discussion_r1488637020

What do we think about doing that? I'm assuming that would be a breaking change? 🤔

PR #759, opened 2025-05-06
Add ember-data to all ember-try scenarios

### LINKED ISSUE #758: `ember-data` v5.3.4+ compatibility
I just updated an app that is using ember-cp-validations to the Ember v5.12 blueprint (coming from v4.12). After the update a runtime error is thrown:

```
Could not find module `ember-data` imported from `(require)` missingModule`.
```
Which is imported here:
https://github.com/adopted-ember-addons/ember-cp-validations/blob/13d8f8457ac9a50370050205bff189f20f526e01/addon/utils/utils.js#L14-L16

I believe this is caused by the fact that `ember-data` [v5.3.4+ is a v2 addon](https://github.com/emberjs/data/commit/1f965a932d907c841ca9629b3788790278ab9ae4), which requires ember-auto-import v2 and ember-cp-validations doesn't list that as a dependency yet.

If I add ember-auto-import v2 to the dependency list of ember-cp-validations the code seems to work again.

Windvis:
Resolves #758  .

Changes proposed:

 -
 -
 -

Tasks:

- [ ] Added test case(s)
- [ ] Updated documentation

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
   "dependencies": {
     "@babel/core": "^7.26.10",
     "@embroider/macros": "~1.16.13",
+    "ember-auto-import": "^2.10.0",
     "ember-cli-babel": "^8.2.0",
     "ember-validators": "^4.1.2"
   },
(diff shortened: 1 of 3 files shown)


Windvis:
Closing since this will take more time than simply changing to a different validation method in our small app. Sorry!

